PR #5, opened 2015-07-23
handle custom fields in schlib

keulYSMB:
This allows to use the same set of key for any field in a schlib.
This way one can create custom library with own fields and default value.


ricardocrudo:
check your pull request running [test_schlib.sh](https://github.com/KiCad/kicad-library-utils/blob/master/schlib/test_schlib.sh).

keulYSMB:
updated. now passing test with any number of fields.

ricardocrudo:
thank you.

PR #7, opened 2015-09-07
add part of check for rule 6.5->Silkscreen is not superposed to pads

andres53016:
Is rustic and need more work, for the moment is funtional for fp_lines

### DIFF
--- a/pcb/rules/rule6_5.py
+++ b/pcb/rules/rule6_5.py
@@ -35,8 +35,14 @@ class Rule(KLCRule):
                 #TODO
                 pass
             elif 'center' in graph:
-                #TODO
-                pass
+                for pad in module.pads:
+                    padComplex=complex(pad['pos']['x'],pad['pos']['y'])
+                    centerComplex=complex(graph['center']['x'],graph['center']['y'])
+                    endComplex=complex(graph['end']['x'],graph['end']['y'])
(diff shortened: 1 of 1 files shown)


michal777:
Works great. Would it be possible to check silkscreen clearance? I've seen in some document about ipc7351 a rule that distance between pad and line on silkscreen should be equal line width (0.15 in case of kicad library). This rule isn't part of KLC but seems to be wise because of manufacturing tolerances. I guess it may be achieved by increasing pad dimensions for calculations.
Anyway let's wait for author of this script.

andres53016:
I add the fix for this comprobation, I have to tidying the script and commit this.
@michal777  you are right, I can add 0.15 to the dimension of pad.
Sorry for my english.

ricardocrudo:
Great! Sorry for delay. Could explain a little bit how the algorithm works? Thank you!

andres53016:
The script translate de fp_line to the origin and translate the same distance the pad, then the vector rotates clockwise the angle with the vector (1,0) and if the pad is  between 0 and the length of the vector, the I have an intersection.
Sorry for my english. I don't know if I explain well.
For the circles is easy. If the distances between center of circunferences is > (radius of pad + radius of circle) then we have an intersection. For the arc not implemented yet, but is similar except that we have to add checking if the pad is between angle start and end of the arc

PR #10, opened 2015-11-16
Add a fix-fields.py script

CarlPoirier:
For now, it just corrects the field text sizes so that the library is compliant with rule 3.6 of the KLC 1.0.


CarlPoirier:
It now reduces pin text and number sizes to 50mils if they are bigger.

ricardocrudo:
Thanks @CarlPoirier

Issue #11, opened 2015-12-12
invalid Violating Rule 6.5 message


pointhi:
it seems like the ```check_kicad_mod.py``` has a little bug with the orientation of square pads, because it throws an invalid error for some of my footprints, where no 6.5 violation is visible.

Furthermore, when using ```--fix``` it delets silkscreen, which would be overlapping the pad, when the pad would be orientated 90 degrees inclined, but not how I have inclined it in the footprint.

### Footprint before --fix (purporsed violation of rule 6.5)

![pad_before](https://cloud.githubusercontent.com/assets/1652144/11763849/32dd8e0e-a11b-11e5-9daa-cccf048b643c.png)

## Footprint after --fix
![pad_after](https://cloud.githubusercontent.com/assets/1652144/11763850/37a30cde-a11b-11e5-90fb-89a037dfeb85.png)

Footprints can be found in this PR: https://github.com/KiCad/RF_Modules.pretty/pull/1

ricardocrudo:
@andres53016 can you check it, please?

andres53016:
Ok! I fix the bug and I create new PR

twam:
Encountered the same bug here, the provided bug fix solves the issue.

ricardocrudo:
Fixed.

pointhi:
Hm, I don't know why, but now many other footprints throw 6.5 Errors for me.

For example the MC9A12 footprints: https://github.com/KiCad/Connectors_Multicomp.pretty/pull/1):
```
checking module: Multicomp_MC9A12_02x32x2.54mm_Straight_MC9A12-6434
  Violating Rule 6.5
    Silkscreen is not superposed to pads, its outline is completely visible after board assembly, uses
    0.15mm line width and provides a reference mark for pin 1. (IPC-7351).
```

The old script doesn't throw errors for those footprints.

ricardocrudo:
@andres53016  Can you check that when you got some spare time?

andres53016:
Hi @ricardocrudo  
It was right on that now.

CarlPoirier:
New PR is https://github.com/KiCad/kicad-library-utils/pull/20

CarlPoirier:
I merged the fix for the rule 6.5 false positive.

PR #12, opened 2015-12-15
some bug fixed in rule 6.5

andres53016:
fixed the bugs mentioned in issue #11


ricardocrudo:
I've merged it with some small changes. I did edit to have same coding style as other files and I renamed one variable to have English name instead of Spanish.

Thank for contributing.

PR #13, opened 2016-01-09
Parse numbers with prefixed +

BenediktSeidl:
pcbnew seems to accept things like

    (fp_line (start 0 4) (end +0.5 4.5) (layer F.SilkS) (width 0.15))

so `check_kicad_mod.py` should too?

`test_kicad_mod.sh` on the other hand will fail for such a file.


ricardocrudo:
Nice. Thank you.

PR #14, opened 2016-02-16
Added windows batch script for checking all .kicad_mod files in a dir

SchrodingersGat:


pointhi:
Does Windows not supporting wildcards?

like
```
./check_kicad_mod.py ../../clones/Connectors_JST.pretty/*.kicad_mod
```

SchrodingersGat:
Apparently not

pointhi:
Probably we can patch it into the script using https://docs.python.org/2/library/glob.html

PR #15, opened 2016-02-29
STM32 library generator

ObKo:
Here is script, which was used to generate stm32.lib for https://github.com/KiCad/kicad-library/pull/304 and https://github.com/KiCad/kicad-library/pull/208

Script was originally written by @DrLuke and improved by me.

This script turns XML files (from ST's STM32CubeMX software) defining pins for STM32 devices into KiCAD library. It also grabs some information (flash size, ram size, speed) from provided datasheets. 

Additional information can be found in README.md


CarlPoirier:
Hi @ObKo,

May I suggest the path schlib/autogen/stm32/ ? Then it's ready to go.

ObKo:
@CarlPoirier, done.

ricardocrudo:
Thanks @ObKo !

SchrodingersGat:
Hi @ObKo  what are your thoughts on limiting the number of alternative pin functions that are displayed? The schematic symbols are very wide when all the pin functions are displayed.

DrLuke:
@SchrodingersGat And how do you decice which functions to leave out? What if someone else wants those alt functions shown on the pins?  

Leaving out functions on the pins wouldn't result in a very good library in my opinion. If the symbols are too wide, you can modify them for a local copy.

SchrodingersGat:
@DrLuke you are correct, it is useful to know all the options. I guess the only other option would be to simply display pin names (PA1, PA2, etc). (which is less preferable to having them all displayed and being a bit large on the schematic)

SchrodingersGat:
@ObKo disregard my last point. However, it does seem that the generated footprints are missing exposed-pad pins for the QFN chips?

SchrodingersGat:
untitled

e.g. STM32F302K6Ux

ObKo:
@SchrodingersGat 

Again, the problem is in invalid ST's pin files -  some of them has no exposed-pad mark for QFN.

I'll add fix to script in this week.

PR #16, opened 2016-02-29
Added options to checklib.py to help with automation of checks

reportingsjr:
Hi,

I made these changes a while back for the github pull request checker.

The three changes are:
1) a flag to not print out violations (the PR checker doesn't need all 1000 violations that the library has to be printed out)
2) a flag to store all errors in a json format. This is so all of the violations the library has can be removed when checking the pull request.
3) Lastly a flag to read in the previous errors in json and ignore those errors.

### DIFF
--- a/schlib/checklib.py
+++ b/schlib/checklib.py
@@ -2,6 +2,7 @@
 # -*- coding: utf-8 -*-
 
 import argparse
+import json
 from schlib import *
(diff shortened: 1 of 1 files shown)


reportingsjr:
Actually, this needs some more work before it should be merged. Ignore this for now!

PR #19, opened 2016-03-06
Improve verbose output

### LINKED ISSUE #17: [Enhancement] checklib structure
I was browsing sources and noticed, that verbose messages are generated in checklib.py script. This is kind of limiting, because each rule script returns just T/F. All additional info is specific for each rule and checklib.py must handle message for each rule separately.
I suggest:
- add *verbose level argument* into rule constructor (rule.py)
- add *list of messages* which will buffer verbose message from rule script (rule.py)
- add method for appending separate verbose messages into *message list* ^^ (rule.py)
- modify all rule scripts to compose verbose messages using that method^^
- modify checklib.py to pass verbose level to rule scripts and print message buffers after check() (this will be unified for all rules, so there will be no need to modify checklib.py for each rule script)

I can do this and give you PR. I just need to know, if you agree with this idea.

diggit:
Improved way how to generate verbose output.
Now each rule script fills its own message buffer with success/info/warning/error messages. checklib.py then selects which to print and how to colorize them. (each message has verbosity and severity flag)
All scripts have message for each detected problem. So no more guessing what could be wrong.

Scripts were tested on KiCad library.

These changes makes a lot of variables unused. Then could be removed in future.

This PR resolves #17

### DIFF
--- a/schlib/checklib.py
+++ b/schlib/checklib.py
@@ -6,6 +6,21 @@ from schlib import *
 from print_color import *
 from rules import *
 
+def processVerboseOutput(messageBuffer):
+    if args.verbose:
+        for msg in messageBuffer:
+            if msg[1] <= args.verbose:
+                if msg[2]==0:#Severity.INFO
+                    printer.gray(msg[0], indentation=4)
+                elif msg[2]==1:#Severity.WARNING
+                    printer.brown(msg[0], indentation=4)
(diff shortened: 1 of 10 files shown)


ricardocrudo:
Great refactoring! It give a boost on user feedback. Thank you!

PR #25, opened 2016-03-29
Vizualising variant BOMs

jnavila:
Variant BOM can be visualized in schematics with the two scripts introduced in this PR.

A first script in schlib `create_nopop.py`, can be applied to the -cache.lib library to create duplicated parts with a big cross on them. These new parts'names are appended with "_NOPOP".

Then, you can apply the second script on the sch files. This script will change the part's symbol to nopop if a given field matches a given value, eg "NP". If you have defined some fields named according to the variant BOM and filled with the given "NP" value when the part should not be populated in this variant, the new schematic file will reflect the variant with crossed parts when not populated.

Be careful that these scripts are "destructive". You should have backed up the project or commited the latest version in you favorite version control system.

### DIFF
--- a/sch/variant.py
+++ b/sch/variant.py
@@ -8,7 +8,11 @@ def variant(mysch, key_field, nopop_flag="NP"):
     for comp in mysch.components:
         fields = [ f['ref'] for f in comp.fields if f['name'] == kf]
         if len(fields)>0 and fields[0] == npop:
-            comp.labels["name"] = comp.labels["name"] + "_NOPOP"
+            if not comp.labels["name"].endswith("_NOPOP"):
+                comp.labels["name"] = comp.labels["name"] + "_NOPOP"
+        else:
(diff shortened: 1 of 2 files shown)


poeschlr:
Sorry this got missed so long.

I am not really sure the library utils repo is the correct lib. The library utils repo is meant to hold stuff usefull for managing the official kicad lib. (Main purpose is the KLC checking part of it.)

It is not really meant to help managing projects. A general scripting repository would be an option. Not sure who should manage that though.

If your script does use infrastructure that already exists in this repo (Example some classes) without changing the general behavior we can merge it. But be prepared that changes to our tools might break yours.

jnavila:
Thanks for taking some time to clarify on this. Sorry, I wasn't aware that this repo was dedicated to the maintenance of the official kicad lib.

The new files rely on this lib for the parsing of library and schematic files. So, I guess my fork will be kept as is.

PR #28, opened 2016-06-15
Windows fixes

SchrodingersGat:
This PR makes some subtle changes to the checklib.py and check_kicad_mod.py scripts to allow them to work better under windows.

* Enable wildcard support - Windows does not have command-line wildcard support by default.
* Fixed reference to *print_color.py* in *check_kicad_mod.py* - Python under windows doesn't support running external scripts in this fashion.

Please check that both of these scripts still perform as intended under *nix


CarlPoirier:
Looks good to me and works fine on Linux.

Thanks!

PR #29, opened 2016-06-16
Add extra check EC05 for detecting duplicate pins

SchrodingersGat:
This PR proposes a new rule for schematic symbols (checklib.py) which checks for duplicated pins in a schematic symbol.

A lot of PRs in /kicad-library have duplicated pins, which is not caught by current rule-set.

Example output: (with full verbosity -vv)

```
checking component: TPS2513
  Violating Rule 3.9
    Pin numbers should not be duplicated for a symbol.
    Pin 1 is duplicated
    Pin 'Gnd' is a duplicate of pin 'DP1'
```


ricardocrudo:
Hello Oliver.

Officially, there is no rule 3.9. In this case I recommend to name it as EC (Extra Check).

SchrodingersGat:
I think we should add this as a rule to https://github.com/KiCad/kicad-library/blob/master/KiCad_Library_Convention.txt as having duplicate pins is a pretty serious error.

SchrodingersGat:
@ricardocrudo I have now made this into an EC rule (EC05)

diggit:
EC05 seems ok to me.

> Pin numbers should not be duplicated for a symbol.

```if p['num'] == pin['num'] and not p['name'] == pin['name']:``` What if numbers are same and names too? I think this is problem of pin numbers only.

IMO output with groups of duplicate pins would be clearer. Sth. like:
```
Multiple occurrence of pin number 1:
   pin: A (1), posx 350, posy -200
   pin: T (1), posx 350, posy -100
   pin: T (1), posx 350, posy 0
```
You can just copy pin line from rule rule3_6.py

SchrodingersGat:
Done. See what you think.

diggit:
That's what I meant. Info about position can be useful for parts with hundreds of pins.
If you agree, then what do you think about fix method, that eg. doubles their length to find them easily.
If not, I'll merge your PR now.

SchrodingersGat:
I have added location information as suggested (good idea!) I'm not sure about the proposed fix though.. I think that the precise location information should be enough?

CarlPoirier:
Hi folks,

The check is a great idea, but for naming I'm also in favor of the "extra
check" naming. While it's true it could be an official rule, if we start
listing such obvious ones like this, the KLC will get very long.

I'll let @Patrick merge it since he did the review.

Carl

diggit:
@CarlPoirier it is EC05 now, shouldn't be a problem. BTW, you notified wrong guy.

Location is enough, definitely. (You could use already made location formater, to stick with already used pin format, but this is ok too.)
FIX:
- Individual pins should have equal SEVERITY to leading message about duplication.

Nothing more, almost there!

SchrodingersGat:
Like so?

diggit:
Just SEVERITY, VERBOSITY level should be HIGH to allow only error output without detailed info.
It extends info about problem -> verbosity++, severity same

SchrodingersGat:
Ah ha, I understand. Fixed.

SchrodingersGat:
Any further requirements?

SchrodingersGat:
I have now also added a check for any missing pins as per https://github.com/KiCad/kicad-library-utils/issues/36

SchrodingersGat:
@diggit or @CarlPoirier any further thoughts on this one?

SchrodingersGat:
Fixed conflicts

diggit:
Works as expected. Thanks!
Sorry for long waiting.
Nice work you did, merging.

PR #31, opened 2016-07-05
Extended the functionality of EC01 (testing pin types)

SchrodingersGat:
This PR improves the functionality of extra-checking rule 1 (EC01) used in the checklib.py script:

* Multiple pin types can be tested
* A lot easier to add extra tests

### DIFF
--- a/schlib/rules/EC01.py
+++ b/schlib/rules/EC01.py
@@ -4,12 +4,38 @@ from rules.rule import *
 import re
 
 class Rule(KLCRule):
+
+    #Power Input Pins should be 'W'
+    POWER_INPUTS = ['gnd','vcc','vdd','ground','vbat','vss','vaa','vin','vi']
+    
+    #Power Output Pins should be 'w'
+    POWER_OUTPUTS = []
+    
+    PASSIVE_PINS = []
+    
+    #Input Pins should be "I" or ""
+    INPUT_PINS = ['sdi','clk','clock','~cs','cs']
+    
+    #Output pins should be "O"
+    OUTPUT_PINS = ['sdo',]
+    
+    #Bidirectional pins should be "B"
+    BIDIR_PINS = ['sda',]
+    
+    tests = {
+        "W" : POWER_INPUTS,
+        "w" : POWER_OUTPUTS,
+        "P" : PASSIVE_PINS,
+        "I" : INPUT_PINS,
+        "O" : OUTPUT_PINS,
+        "B" : BIDIR_PINS,
+        }
+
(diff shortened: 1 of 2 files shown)


diggit:
Great idea!
I'd suggest changing list POWER_PINS to POWER_INPUTS, because we have power outputs too and sb. can mix them with power inputs easily. In fact, you can create list for all pin types and comment out unused, as template for future.
Other variable names could be polished too, but not necessarily.

Few more pin names to add:
- **power inputs:** vbat, vi, vss
- **open collector:** scl
- **inputs:** cs (but a lot of false positives)
- possibly other, mostly single purpose pins: xtal, aref, vref,..

SchrodingersGat:
@diggit I have made the following improvements:
* Changed POWER_PINS to POWER_INPUTS
* Added the POWER_OUTPUTS test
* Added the PASSIVE_PINS test
* **Fixed a spelling mistake**
* Added the ability to --fix any pin based on the provided tests

diggit:
Thanks, looks good.
**vin** and **~cs** are redundant, they are matched by **vi** and **cs**.
You can remove them.

diggit:
oh and can you add "not connected" (N) as test for pins NC ?

SchrodingersGat:
@diggit vi/vin and cs/~cs would not have been equivalent as the matching was being performed on a literal basis.

However I have now changed it to implement (case-insensitive) regular expression matching.

I have also added a NC test.

Let me know what you think!

diggit:
I tested it on simple test part and:
```
  Violating EC01 - Extra Checking
    Check pins names against pin types.
    pin GNDA (1): I (INPUT), expected: W (POWER INPUT)
    pin Vin (3): I (INPUT), expected: W (POWER INPUT)
    pin GPIO1/SDO/TXD (8): B (BIDI), expected: O (OUTPUT)
```
Last line in wrong, there are 2 solutions in my mind:
- (simpe) match only whole name -> works for dedicated pins
- (complicated) create more complex algorithm. Build list of required pin capabilities SDO(OUT)+ RXD(IN) -> BIDI, IN+OC=?,... and compare

You can do simple solution. Implementation of complicated one is fun for days with empty PR queue.

SchrodingersGat:
I have added ^ and $ symbols to the regexes such that complete pin names must be checked.

This approach leaves the regex functionality in place (can be made more complicated later on) but means that only simplified checks are being performed.

diggit:
Oh, sorry. I forgot to pull changes.
Nice work you did, merging.

Issue #4, opened 2015-06-26
move_part script doesn't move/remove alias (dcm file)


ricardocrudo:
When a component has alias the respective documentation is not removed from source and it is not placed to destination.

PR #32, opened 2016-07-05
Fix schlib.py andmove_part.py to move components properly

### LINKED ISSUE #30: move_part.py does not move alias documentation
When moving parts with script `move_part.py` symbol and documentation of main symbol is moved. Unfortunately documentation of all aliases remain in source dcm file.

Once this is fixed, checklib.py should be extended to check unused aliases. There could be a lot of them.

diggit:
I made some changes to be able to move components properly without abandoning aliases documentation.

SchLib class nnow has 2 additional methods:
- addComponent( component_reference )
- removeComponent( component_name )
 - returns reference to removed component

Both methods handles symbol + documentation (incl. aliases).

This PR fixes #30 and #4

### DIFF
--- a/schlib/move_part.py
+++ b/schlib/move_part.py
@@ -35,40 +35,12 @@ for comp in dst_lib.components:
         sys.exit(1)
 
 # append component to destination and save
-dst_lib.components.append(component)
+dst_lib.addComponent(component)
 dst_lib.save()
 
 # remove component from source and save
-src_lib.components.remove(component)
+src_lib.removeComponent(component.name)
 src_lib.save()
-
-if component.documentation:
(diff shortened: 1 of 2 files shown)


ricardocrudo:
Great work. I've tested trying to move MAX2606 (maxim.lib) to somewhere else and I got error removing a component that actually is an alias. The error is in schlib.py:95. Adding the following test before delete do the job: `if name in self.components.keys():`

Fixing this it should be ready to merge.

diggit:
Yeah, I forgot on parts without documentation.
Solution works, but it still created empty records in destination library, which KiCad erases. Fixed it too.

ricardocrudo:
Thanks.

PR #33, opened 2016-07-06
Fix parsing Fx fields

diggit:
Until now, any line starting with "F" was parsed as field. Problem occured for footprint list entries starting with "F". They were incorectly parsed as fields.


ricardocrudo:
Thanks Patrik.

PR #34, opened 2016-07-09
Added the ability to select individual rules for checklib.py test

SchrodingersGat:
This PR adds the -r --rule flag to the checklib.py script.

It allows the user to specify particular rules to check against (if not provided, then ALL rules will be checked, as is the default).


diggit:
You read my mind! This is next necessary feature to add.

Two things to improve:
- If you forget _--enable-extra_ parameter and you have eg. EC01 in rules to check against. Result looks like it passed, but check didn't happen at all. Throw warning about requested EC* check without --enable-extra param. (maybe turn it on additionally?)
- user may not know what format of rules script expects KLC3.1, 3.1 rule3.1,...? Add some short example to help like _-r 3.2,EC01,..._

SchrodingersGat:
Good suggestions, implemented.

ricardocrudo:
Nice!

PR #35, opened 2016-07-09
Added the 'silent' option to checklib.py

SchrodingersGat:
This PR adds the -s / --silent flag to checklib.py

If set, the checklib.py script hides any output for components that do not have any KLC violations.

e.g. when checking a large library (or multiple libraries) it can be useful to only display the components that DID violate, rather than having many lines of "No violations found".

Then the -s option is not specified, action is default.

### DIFF
--- a/schlib/checklib.py
+++ b/schlib/checklib.py
@@ -31,7 +31,7 @@ parser.add_argument('--fix', help='fix the violations if possible', action='stor
 parser.add_argument('--nocolor', help='does not use colors to show the output', action='store_true')
 parser.add_argument('--enable-extra', help='enable extra checking', action='store_true')
 parser.add_argument('-v', '--verbose', help='show status of all components and extra information about the violation', action='count')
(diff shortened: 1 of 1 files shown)


diggit:
Good idea.

Can you make help message more brief? Complete sentence is not required.
- _'If the silent option is set, there will be no output displayed for components with zero violations. This option is useful for checking large libraries'_
- _"skip output for symbols passing all checks_"

You probably don't need multiple levels of silent, action='store_true' is enough

SchrodingersGat:
Done!

ricardocrudo:
Good idea. I'm wondering if actually it should be the default behavior and  then create a flag to show the components with no violations.

SchrodingersGat:
@ricardocrudo would you like me to invert the operation of this flag, or merge as-is?

SchrodingersGat:
Anyone have any further comments on this one? I think it is good to merge.

PR #38, opened 2016-08-03
Non-Zero exit code for when check scripts fail to pass

bmv437:
Hi there. So I've added a non-zero exit code when `checklib.py` doesn't pass all checks. This is useful for using `checklib.py` as a pre-commit hook, or for any other automated build/test system.

I have an example of it working as a pre-commit hook here:
https://github.com/bmv437/kaleidoscope_library

Thoughts? I also plan on adding non-zero exit for `check-kicad-mod.py` as well.

### DIFF
--- a/schlib/checklib.py
+++ b/schlib/checklib.py
@@ -2,6 +2,7 @@
 # -*- coding: utf-8 -*-
 
 import argparse
+import sys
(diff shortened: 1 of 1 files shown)


SchrodingersGat:
Great idea :)

PR #39, opened 2016-08-07
Added --silent option for check_kicad_mod.py

SchrodingersGat:
This PR adds the -s option to the kicad_mod checker, and hides output for modules that pass all checks (e.g. when checking an entire directory of footprints)


SchrodingersGat:
Fixed merge issues

PR #43, opened 2016-08-23
Added regex matching to check multiple components in a library

SchrodingersGat:
This PR adds the ability to check multiple components in a library by using a regular expression with the -p command line option


CarlPoirier:
Thank you!

PR #44, opened 2016-08-25
Additions to EC01

SchrodingersGat:
This PR adds two new checks for the EC01 rule
- Added test for missing pin name
- Added test for NC pins that are set to visible


ricardocrudo:
Thanks and sorry for so late reply.

PR #45, opened 2016-08-25
Simple check of KLC  3.7

SchrodingersGat:
This is a very simple checker against KLC 3.7

> 3.7.  The Value field is prefilled with the object name.

Checks that F1 value equals the component name


SchrodingersGat:
Fixed conflicts

ricardocrudo:
Thank you.

PR #46, opened 2016-08-25
RFC: KLC 3.9

SchrodingersGat:
In this PR I propose a new rule for KLC, 3.9, which checks the values of the default footprint for a component.

1. If the footprint field is empty, then the check passes (ignore any further tests)
2. Footprint field should be set to invisible
3. Footprint field should be of the format "Library:Footprint" (i.e. it must have a single ":" character)
4. Footprint field cannot have any illegal path-name characters e.g. \/*?:"<>| 

I think such a test (with a good description in the KLC rule file) complements the current set and will also help anyone wanting to submit parts.

### DIFF
--- a/schlib/rules/__init__.py
+++ b/schlib/rules/__init__.py
@@ -1,2 +1,2 @@
 
-__all__ = ["rule3_1", "rule3_2", "rule3_6", "rule3_8", "EC01", "EC02", "EC03", "EC04"]
+__all__ = ["rule3_1", "rule3_2", "rule3_6", "rule3_8", "rule3_9", "EC01", "EC02", "EC03", "EC04"]
--- a/schlib/rules/__init__.py
+++ b/schlib/rules/__init__.py
@@ -1,2 +1,12 @@
 
-__all__ = ["rule3_1", "rule3_2", "rule3_6", "rule3_8", "rule3_9", "EC01", "EC02", "EC03", "EC04"]
+__all__ = [
+"rule3_1",
+"rule3_2",
+"rule3_6",
+"rule3_8", 
+"rule3_9", 
+"EC01", 
+"EC02", 
+"EC03", 
+"EC04"
+]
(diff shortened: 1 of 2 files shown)


CarlPoirier:
Hi @SchrodingersGat,

Your code is great! The description as a KLC rule must however be reworked.

> Default component footprint is appropriately set.

This means nothing really. It needs to say the contents of points 1, 2, 3 and 4. BTW, wildcards are accepted in the footprint field so that is a valid character as well as the colons for separation.

Carl

SchrodingersGat:
Carl,

I thought that the default footprint had to be explicit, and that the wildcard functionality was for the footprint filters? 

The rule text was a bit of a placeholder, I will fix this.

SchrodingersGat:
Rule description has been fixed.

The KLC rules themselves need some work, I think, as they don't convey all the information that we are actually checking for. In the long-form document, each rule could do with a quick sentence describing it, or a screenshot if that helps.

Which is the "official" KLC out of:

https://github.com/KiCad/kicad-library/wiki/Kicad-Library-Convention
https://github.com/KiCad/kicad-library/blob/master/KiCad_Library_Convention.txt

Is there a reason for this duplication? I think the wiki version should be the master and the .txt file in the repo could be replaced by a link to the wiki page in *Contributing.md*. What are your thoughts on this?

CarlPoirier:
Hi @SchrodingersGat,

I indeed thought you talked about the footprint filters. Since it's optional, shouldn't it be better to put it in the extra checks?

As for the KLC documents, they are supposed to be equivalent. The reason is that the .txt came first and eventually it got added to the wiki.

SchrodingersGat:
I think that filling out the default footprint field (or leaving it blank if there are multiple options) should be a major rule, as this is something that I (and other librarians) deal with a lot - explaining to users submitting new symbols the purpose of the default footprint field.

I also think that it should be an official rule as then we could describe it 

> As for the KLC documents, they are supposed to be equivalent. The reason is that the .txt came first and eventually it got added to the wiki.

So, should we remove KiCad_Library_Convention.txt and just link to the wiki page?

SchrodingersGat:
Carl here's my simple suggestion for the duplicated rule information : https://github.com/KiCad/kicad-library/pull/686/files

CarlPoirier:
I will add rule 3.9 to the next revision of the KLC which should happen this weekend.

PR #48, opened 2016-08-26
.lib comparison

SchrodingersGat:
This PR adds two key features:

**Component Checksum**
The schlib.py script now calculates component-level checksums. This can be used to determine which parts have been updated between different versions of the .lib file

**Library Comparison Script**
The *comparelibs.py* script takes two versions of a .lib file, and lists which components have been added / deleted / changed. If the --check flag is given, any updated/additional components are automagically run through the KLC script

(The -silent option for checklib.py has also been merged into this branch for argument pass-through)

### DIFF
--- a/schlib/checklib.py
+++ b/schlib/checklib.py
@@ -78,7 +78,11 @@ exit_code = 0
 for libfile in libfiles:
     lib = SchLib(libfile)
     n_components = 0
-    printer.purple('library: %s' % libfile)
+    
+    # Print the library name if multiple libraries have been passed
+    if len(libfiles) > 1:
+        printer.purple('library: %s' % libfile)
+    
     for component in lib.components:
         # skip components with non matching names
         if args.component and args.component != component.name: continue
--- a/schlib/comparelibs.py
+++ b/schlib/comparelibs.py
@@ -23,11 +23,12 @@ parser.add_argument("--check", help="Perform KLC check on updated/added componen
 args = parser.parse_args()
 
 def KLCCheck(component):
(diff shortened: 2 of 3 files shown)


SchrodingersGat:
Fixed conflicts

SchrodingersGat:
I have now reduced the verbosity level of the KLC call, and added a return code for the comparison script.

CarlPoirier:
This is awesome!

jacobrosenthal:
Late thought on this. How about output being still color coded but with + - in front like a diff instead of the human readable. I think this is easier to parse by the next tool
```
-ACS712ELCTR-05B-T
-ACS711ELCTR-12AB
-ACS722LLCTR-05AB-T
-ACS754SCB-200
-ACS706ELC-05C
+ACS706ELC-WHATEVER1
+ACS706ELC-WHATEVER2
```

jacobrosenthal:
Seems like youll want to probably do more of this in python yet, but Ill still keep problem solving in bash

Playing with this script https://github.com/jacobrosenthal/kicad-library/blob/master/test.sh to wget changed files and throw them at the new comparelibs.py

I posted here https://github.com/KiCad/kicad-library-utils/pull/48#issuecomment-243970675 but I think we could make comparelibs.py more parseable for the next step with + and - signs for components changed.

jacobrosenthal:
Wrong window. Sorry for noise. Deleted that comment

SchrodingersGat:
See https://github.com/KiCad/kicad-library-utils/pull/50 

Both suggestions have been implemented.

PR #50, opened 2016-09-01
Improved output of comparelibs.py

SchrodingersGat:
Updated output of the script to make it easier to parse with a higher level script as per https://github.com/KiCad/kicad-library-utils/pull/48


CarlPoirier:
Great, thanks!

PR #65, opened 2016-11-09
Decode library text in schlib.py

piotr-wiszowaty:
This is to avoid problems with encoding non-ASCII characters.

### DIFF
--- a/schlib/schlib.py
+++ b/schlib/schlib.py
@@ -133,7 +133,7 @@ class Component(object):
         checksum_data = ''
         
         for line in data:
-            checksum_data += line
+            checksum_data += line.decode('utf-8')
             line = line.replace('\n', '')
             s = shlex.shlex(line)
             s.whitespace_split = True


ricardocrudo:
hello @piotr-wiszowaty can you please point me an example which the decoding is necessary in this case?
This change you are suggesting does not pass in the test script: test_schlib.sh

piotr-wiszowaty:
You mean that test_schlib.sh detects differences in outputs when testing stm32.lib and transistors.lib?
The result is similar with or without the above patch.
The problem with encoding on my machine (master branch) looks like this:
```
$ checklib.py -v --enable-extra motorola.lib
  File "/home/p/bin/checklib.py", line 82, in <module>
    lib = SchLib(libfile)
  File "/home/p/src/kicad-library-utils/schlib/schlib.py", line 296, in __init__
    self.__parse()
  File "/home/p/src/kicad-library-utils/schlib/schlib.py", line 328, in __parse
    self.components.append(Component(component_data, comments, self.documentation))
  File "/home/p/src/kicad-library-utils/schlib/schlib.py", line 220, in __init__
    x = checksum_data.encode('utf-8')
UnicodeDecodeError: 'ascii' codec can't decode byte 0xc2 in position 54921: ordinal not in range(128)
$ uname -a
Linux t.local 4.8.7-200.fc24.x86_64 #1 SMP Fri Nov 11 15:44:18 UTC 2016 x86_64 x86_64 x86_64 GNU/Linux
$ python -V
Python 2.7.12

```

ricardocrudo:
Fixed. Thanks.

PR #68, opened 2016-11-27
Make this library "installable" with pip, added component editor class

oyagev:
1. Added `__init__.py` files to directories, to make python see those as packages
2. Added `setup.py` to allow installation by pip. I left the author field empty. 
3. Added Editor class to allow direct editing of component fields. 
4. I suggest adding the library to PyPI
5. I suggest separating the library / classes from the python scripts. Either to different "scripts" folder or even another repository.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,1 +1,4 @@
 *.pyc
+KicadLibUtils.egg-info/
+.idea/
+
--- a/.idea/vcs.xml
+++ b/.idea/vcs.xml
@@ -1,6 +1,0 @@
-<?xml version="1.0" encoding="UTF-8"?>
-<project version="4">
-  <component name="VcsDirectoryMappings">
-    <mapping directory="$PROJECT_DIR$" vcs="Git" />
-  </component>
-</project>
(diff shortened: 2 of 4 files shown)


Ratfink:
Sorry this didn't get any attention sooner!  I'd personally love to see this repository refactored so that it can be treated as a big package (or possibly a set of packages), put on the [PyPI](https://pypi.python.org/pypi), and installed with `pip`, but I'm afraid this would be a major task requiring a lot more changes than this PR includes.

* Many or all of the scripts would have to be rewritten as executable modules or packages.
* All the packages (`pcb`, `schlib`, etc.) should be either made as sub-packages of one big package, or should be split into separate git repositories and packaged independently.  I could argue either way.
* The `schlib/autogen/KiCadSymbolGenerator` package would need to be fixed up a bit, and maybe be merged with the `schlib/schlib.py` module to make a single representation of the schematic libraries (this might not be worth doing before the .sweet format is finalized).
* Everything that has a public API would need detailed documentation.
* Maybe more stuff I haven't thought of.

Because it would be such a major undertaking, I don't think we should do it immediately, and the changes in this PR don't seem like the right starting point.  Unless other maintainers object, I'll plan to close this PR in a few days and open an issue to discuss what, if anything, we should do on this front.

evanshultz:
Great idea! Right now, though, I agree with the direction @Ratfink proposes.

PR #69, opened 2016-11-30
Fixes for klc v1.1

openelectron:
Edited rules 6.5 and 6.6 in accordance with the updates to the KLC v1.1


CarlPoirier:
Thank you very much!

PR #71, opened 2016-12-25
(KLC 6.6) check if end of line is aligned on 0.01mm grid instead of 0.05mm

chrisnoisel:
I noticed the the test checked if end of the lines of the courtyard were on a 0.05mm grid instead of 0.01mm (KLC 6.6).


SchrodingersGat:
@chrisnoisel thanks for this - it has been addressed in https://github.com/KiCad/kicad-library-utils/pull/74

PR #72, opened 2016-12-26
Fix rule 6.9

piotr-wiszowaty:
1. Update thickness to match KLC
2. Update rule message to actually match the checks done by the script

### DIFF
--- a/pcb/rules/__init__.py
+++ b/pcb/rules/__init__.py
@@ -1,3 +1,3 @@
 
-__all__ = ["rule6_3", "rule6_4", "rule6_5", "rule6_6", "rule6_9",
+__all__ = ["rule6_3", "rule6_4", "rule6_5", "rule6_6", "rule6_9", "rule6_10",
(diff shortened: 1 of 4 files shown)


SchrodingersGat:
@piotr-wiszowaty I am a little confused by your PR here - you reference a KLC6.10 but there is no such rule?

Additionally, these fixes are addressed in https://github.com/KiCad/Pin_Headers.pretty/pull/17, so unless you can see something that is missing from @jkriege2 's PR I will go with that one.

piotr-wiszowaty:
On 12/27/2016 11:46 PM, Oliver wrote:
>
> @piotr-wiszowaty <https://github.com/piotr-wiszowaty> I am a little 
> confused by your PR here - you reference a KLC6.10 but there is no 
> such rule?
>
Right - my mistake.

> Additionally, these fixes are addressed in KiCad/Pin_Headers.pretty#17 
> <https://github.com/KiCad/Pin_Headers.pretty/pull/17>, so unless you 
> can see something that is missing from @jkriege2 
> <https://github.com/jkriege2> 's PR I will go with that one.
>
In that case I'm closing mine.

PR #73, opened 2016-12-26
SQLite database driven footprint assignment

leventelist:
This was originally developed for gEDA, and ported to KiCad.

With this tool you can have a database with devices. The script looks for ID=xxx attribute, and adds/replaces footprint of the component. The database can be used to monitor stock as well.

Apart from the footprint, other attributes can be replaced too.

### DIFF
--- a/sch/db/db.sql
+++ b/sch/db/db.sql
@@ -1,0 +1,64 @@
+BEGIN TRANSACTION;
+CREATE TABLE supplier (
+    id INTEGER PRIMARY KEY AUTOINCREMENT,
+    name TEXT,
+    homepage TEXT,
+    address TEXT,
+    country TEXT,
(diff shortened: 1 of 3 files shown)


SchrodingersGat:
Hi @leventelist this looks like it would be a useful tool, but I do not think it is appropriate to add to this set of tools. The kicad-library-utils are for back-end library management, which are (generally) not seen by end-users of KiCad.

Your scripts would be much more suitable as a stand-alone tool for KiCad users to make use of in their projects.

I would suggest developing this as a separate tool  which could then be featured on the [External Tools](http://kicad-pcb.org/external-tools/) page.

leventelist:
Okay. I close this pull request, and put it elsewhere. Thanks for the comment.

PR #74, opened 2016-12-27
improved kicad_mod check scripts:

jkriege2:
1. added additional messages in verbose-mode that explain WHY a rule failed
  2. removed old rule 6.9 and redistributed to 10.4 and 10.5 (compare KLC on Webpage)
  3. defined required widths/thicknesses/grids as variables so they can be changed centrally
  4. things that need to be on FAB/SILKS-layer may be either F.Fab/SILKS or B.Fab/SILKS
  5. fixed bugs in rule 6.6:
       - that lead to false-positives (%INT instead of %FLOAT)
       - that required different grids for start/end of line
       - fixed fix()


pcb/rules/rule10_4.py
```diff
@@ -7,14 +7,34 @@ class Rule(KLCRule):
...
         """
         Proceeds the checking of the rule.
         """
         module = self.module
-        return (module.value['value'] != module.name or module.value['layer'] != 'F.Fab')
+        ok=False
+        if module.value['value'] != module.name:
+            self.verbose_message=self.verbose_message+"Contents of value label ('{0}') is different from module name '{1}'!\n".format(module.value['value'], module.name)
+            ok=True
+        if module.value['layer'] != 'F.Fab' and module.value['layer'] != 'B.Fab':
```
SchrodingersGat: Suggestion: 
```if module.value['layer'] not in ['F.Fab', 'B.Fab']:```

SchrodingersGat:
This is a very thorough update. I would like to see a simple test for KLC6.9 (even just testing for presence of F.Fab line)

SchrodingersGat:
i would like to see a check for the *new* KLC6.9, even if it is just testing for the presence of a F.Fab line of the appropriate width.

jkriege2:
There you go ;-)

Best,
JAN

SchrodingersGat:
Ok, this looks great as far as I can tell :)

PR #77, opened 2017-01-17
fixes/improvements to rules 6.3, 6.6 and 10.3 + commandline-options --addsilkscreenrect and --rotate

jkriege2:
- rule 10.3: now also replaces ';' as separator
- rule 10.5: now adds a user-text '%R' on F/B.Fab so the REFDES is visible on F.Fab AND F.SilkS (see new KLC)
- rule 6.6: 
    - now `--fix` will add a rectangular courtyard with 0.2mm offset to the pads and F.Fab (or F.SilkS if F.Fab does not exist) and if no CrtYd exists so far
    - now this rule can `--fix` a too-large/too-small courtyard rectangle. If the courtyard contains no rectangle with vertical7horizontal lines only it is not checked/altered (e.g. for packages with round courtyards, or advanced shapes)
- rule 6.3 now does not crash if pin1 does not exist + it tries to find A1 if 1 was not found + it now also moves the 3D model
- added rule 1.8, which checks for UNIX-style line-endings + `kicad_mod`-files are now explicitly saved with UNIX-style line-endings
- added a commandline-option `--addsilkscreenrect` that simply adds a silkscreen rectangle around what's found in the footprint (F.Fab, pads, ...)
- added a commandline-option `--rotate degrees` that rotates the full footprint by a given number of degrees (clock-wise), also rotates the 3D model correctly!

### DIFF
--- a/pcb/kicad_mod.py
+++ b/pcb/kicad_mod.py
@@ -3,6 +3,18 @@
 
 import sexpr, re, math
 
+def _rotatePoint(x, degrees):
+	xx=x        
+	y={'x':x['x'], 'y':x['y']}
+	y['x']=math.cos(degrees/180.0*math.pi)*xx['x']+math.sin(degrees/180.0*math.pi)*xx['y'];
+	y['y']=-math.sin(degrees/180.0*math.pi)*xx['x']+math.cos(degrees/180.0*math.pi)*xx['y'];
+	if 'orientation' in x:
+		y['orientation']=xx['orientation']-degrees               
+	if 'z' in x:
+		y['z']=xx['z']
+	#print('before rotation (degrees=', degrees, ') x=', xx, '    after rotation y=', y)
+	return y
+
 class KicadMod(object):
     """
     A class to parse kicad_mod files format of the KiCad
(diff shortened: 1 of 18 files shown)


jkriege2:
@SchrodingersGat could you merge this? I think the fixes/tools herein are very handy ;-) ... used it to e.g. fix/rotate the footprints in https://github.com/KiCad/TO_SOT_Packages_SMD.pretty/pull/33

JAN

pcb/kicad_mod.py
```diff
@@ -1,14 +1,32 @@
...
     """
     A class to parse kicad_mod files format of the KiCad
     """
     def __init__(self, filename):
         self.filename = filename
+        
+        # check file line-endings
+        self.unix_line_endings=True
+        filecontentsraw=open(filename,"rb").readline()
+        if (filecontentsraw[-2]==0x0D and filecontentsraw[-1]==0x0A) or (filecontentsraw[-1]==0x0D):
```
SchrodingersGat: if filecontentsraw.endswith("\r\n"):

SchrodingersGat:
Jan these are some good improvements - please see the comments I have made in the code review above

pcb/rules/rule1_8.py
```diff
@@ -0,0 +1,33 @@
...
+import re, os, math
+
+class Rule(KLCRule):
+    """
+    Create the methods check and fix to use with the kicad_mod files.
+    """
+    def __init__(self, module):
+        self.expected_width=0.05
+        self.expected_grid=0.01
+        super(Rule, self).__init__(module, 'Rule 1.8', "Library- and footprint-files use UNIX-sytle line-endings (i.e. \\n = LF = 0x0A).")
```
SchrodingersGat: Possibly rephrase to;
`Library files should use UNIX-style line-endings (LF)`

jkriege2:
Great!

will let you know, when I'm done ...

JAN

pcb/kicad_mod.py
jkriege2: I tried that at first, but it didn't work ... so I will leave my version

pcb/rules/rule1_8.py
jkriege2: Done!

jkriege2:
OK, one change done, the other does not work (see comments above)!
JAN

jkriege2:
@SchrodingersGat What shall we do with the fix courtyard tool? I rather like it, as it automatizes a lot of nasty stuff with the courtyards (at least as long as they are rectangular), but I understand it leads to a lot of alarms due to small deviation of a fe 1/10-1/100th millimeter. Maybe we should activate that as an extra with a commandline-option? E.g. `--proposeandifxcourtyard`?

JAN

jkriege2:
Hi @SchrodingersGat  !
I added two new features (as discussed elsewhere):
1. a command-line option `--fixmore` allows to enable additional fixes. Currently fixing the courtyard if a wrong xlearance was used is the only thing enabled in addition by this. Also I reworked the error message for the latter thing to now contain the used and expected clearances instead of the rectangle, which was hard to read. the expected crt-rect is given in a separate line as a hint
2. I added a variable to the Rule-class which allows the fix()es to print their own verbose messages.

Can we merge this now?

Best,
JAN

SchrodingersGat:
Jan, thanks for all these changes, sorry that I've been so busy.

I will merge, and then try to merge these changes into the KLC rearranging.

Thanks!

PR #79, opened 2017-01-19
Added script to download all .pretty libs (master)

SchrodingersGat:
This PR adds a python script that can download all (or some) of the .pretty libs from the KiCad github account.

## Features
* Uses the latest-copy of fp-lib-table to determine which libs to download (downloads from kicad/kicad-library)
* Ignores libs marked as "deprecated"
* `-d` flag can be used to force download of deprecated libs
* Default action = `git clone`
* `-s | --static` flag switches to download a .zip file of the latest master
* Allows regular expression matching to include or exclude libs
* Update libs with the `-u` flag
* Specify different download dir with `-p` flag

## Requests
I have tested this on Windows, Python3.x, and (partially) on Python2.x. Testing this on *nix would be muchly appreciated.

## Requirements
* Python
* Internet connection (duh)
* git installed in PATH (unless downloading static copies)
* ability to type things into command prompt

@pointhi @diggit @CarlPoirier @jkriege2 @ashtonchase if any of you guys want to take a look and see if you can break it?

Thanks :)


SchrodingersGat:
( A matching script `pretty_lib_updater.py` will be available soon)

ricardocrudo:
Hello @SchrodingersGat 
There is a script available to do such thing at KiCad source code. Check out the link in the README of https://github.com/KiCad/kicad-library.

SchrodingersGat:
Hi @ricardocrudo - if you are referring to e.g. http://bazaar.launchpad.net/~kicad-product-committers/kicad/product/view/head:/scripts/library-repos-install.bat

The scripts in this PR are submitted as an update to the existing scripts, with the following key improvements:

### Drawbacks of current scripts
1. The scripts in the source code are very out of date (with regard to *which* .pretty repositories are included). 
1. The existing scripts have hard-coded references to the various .pretty libs
1. No way of excluding libs (e.g. because deprecated or unwanted)
1. Different scripts required for Windows / Linux / OSX
1. No facility to keep libraries up-to-date

### Advantages of new scripts
1. The *new* scripts download the latest *fp-lib-table* file (not hardcoded) and thus do not need to be kept in line with the libraries.
1. Can batch-update libraries from Github
1. Can automatically exclude libs marked as deprecated
1. Ability to exclude or select particular libs (based on regex matching)
1. Single script for all OS targets

I hope these advantages make sense.

CarlPoirier:
This is nice, but it does not work.

1. Getting the fp-lib-table won't work in Python2 because of urlrequest. I suggest you simply drop Python2 support.
2. There is a problem running the git clone command. I think you need to specify the 'clone' and the url as command arguments instead of one whole call.

SchrodingersGat:
@CarlPoirier thanks for the feedback. I've only tested on Windows and Python3. Currently in Ubuntu so I'll robustify the script and report back.

SchrodingersGat:
@CarlPoirier I have tested and fixed both these issues. Please let me know if you still see issues.

CarlPoirier:
I see two more issues.
1. The -s option errors out on download
2. The -u option fails when getting to a library which has not been cloned yet. This happens with deprecated libraries unless the -d option was ran before.

SchrodingersGat:
Thanks, I have removed the -s option (was really just for my testing) and fixed the -u option

CarlPoirier:
Works great now. Thanks!

PR #80, opened 2017-02-07
Add initial test script to check footprint Pull Requests

pointhi:
This is some intial script, to support travis on .pretty repositories. It looks what files where changed, and runs ```check_kicad_mod.py``` on it.

Example ```.travis.yml```:
```yml
sudo: false
language: python
python:
  - "3.6"
install:
 - (cd /tmp; git clone https://github.com/pointhi/kicad-library-utils.git -b ci-impl-1)
script:
- /tmp/kicad-library-utils/pcb/test_travis_changes.sh
```

You can try the implementation on a test repo I made:

https://github.com/pointhi/Footprint_Test_Repo.pretty

@SchrodingersGat @jkriege2 @CarlPoirier .... What do you think about this idea?

### DIFF
--- a/pcb/test_travis_changes.sh
+++ b/pcb/test_travis_changes.sh
@@ -1,0 +1,51 @@
+#!/bin/bash
+
+# This script file is used to check for error in Pull Requests made on Github
+# it loads a list of changed files, and run check_kicad_mod.py on it
+
+PWD_DIR="$( cd "$( dirname "${BASH_SOURCE[0]}" )" && pwd )"
+CHECK_KICAD_MOD="$PWD_DIR/check_kicad_mod.py"
+
+# https://docs.travis-ci.com/user/environment-variables/#Default-Environment-Variables
+
+if [ -z ${TRAVIS_COMMIT_RANGE+x} ]
+then
+	echo -e "\033[0;31m * env variable 'TRAVIS_COMMIT_RANGE' not set\033[0m"
+	exit 2;
+fi
+
+echo " * check libraries modified in commit range: $TRAVIS_COMMIT_RANGE"
+
+MODIFIED_LIBARIES=`git diff --stat "$TRAVIS_COMMIT_RANGE" | grep -oP '^[[:space:]]*\K.*\.kicad_mod'`
+
+if [ "" == "$MODIFIED_LIBARIES" ]
+then
+	echo " * no .kicad_mod files were changed"
+	exit 0 # nothing to do
+fi
+
+echo
+echo " * found some .kicad_mod files which changed:"
+echo "$MODIFIED_LIBARIES"
+
+echo
+echo " * run '$CHECK_KICAD_MOD' for those files"
+
+$CHECK_KICAD_MOD $MODIFIED_LIBARIES
+
+if [ $? -eq 0 ]
+then
+	echo
+	echo -e "\033[0;32m * no errors found\033[0m"
+	exit 0; # no errors found
+else
+	echo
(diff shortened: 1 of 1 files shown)


jkriege2:
Hi1

nice! I think that would really help us ... I have some wuestions though:
1. Could you run the check-script with the option `-v` to get more explanations?
2. is it possible to simply ignore the checks and still merge (sometimes there are reasons to not follow KLC for one or the other symbol ... or even a bug in the check-scripts)
3. Would it be possible to provide the use a button that runs `--fix` on the commit?
4. Would it be possible to have the output of the check-script directly (not the full log) on the page of teh PR directly (or e.g. as an automatically generated comment)?

best,
JAN

pointhi:
> is it possible to simply ignore the checks and still merge (sometimes there are reasons to not follow KLC for one or the other symbol ... or even a bug in the check-scripts)

That's something which can be defined in the repository settings. So yes, we can merge with failing tests

> Would it be possible to provide the use a button that runs --fix on the commit?

I don't know about this possiblity. And I think we shouldn't encourage it, because this command could probably cause some bugs on the footprint (and wrong formating of the file)

> Would it be possible to have the output of the check-script directly (not the full log) on the page of the PR directly (or e.g. as an automatically generated comment)?

Should be possible, but likely requires some more work and specific setup work (allowing api communication). I think the travis output should be enough at the moment, and we can add this feature later

SchrodingersGat:
I really like this idea but do not think the --fix flag should be implemented here. This should still require explicit user action.

I'm currently away but hope to have a look at this on Sunday or Monday

pointhi:
here is some example output, how a failed PR would look like in travis:

https://travis-ci.org/pointhi/Footprint_Test_Repo.pretty/builds/199033875

pointhi:
Some little ideas which would probably also useful:
* run whole repository check on push, to check KLC compliance of whole repo (which would allow us to quickly check for full KLC-compilcance).
* run the repository checks on a weekly basis, to get notifications when there were changes on KLC / test script

jkriege2:
Hi!

I agree to all above ;-)

Some additional ideas:
1. I would surround the actual check-script-run by some elemnt that draws the attention to it, e.g. 

> ================================================================
> == BUG REPORTS/TEST RESULTS/...
> ================================================================

Also I would add the note that directs to the script +KLC at the top below this marker... and extend it a bit:

>Footprints for the official KiCAD library have to follow the KiCAD library conventions (KLC):
>   https://github.com/KiCad/kicad-library/wiki/Kicad-Library-Convention
>
> There is also a Python script which simplify checks of most KLC rules:
>   kicad-library-utils (https://github.com/KiCad/kicad-library-utils)
>running this script with `-v --fix` may even help to fix several types of rule violations.
>
> CHECK OUTPUT:
> ...[here comes the check output ...] ...

What do you think.
JAN

SchrodingersGat:
I agree with Jan's points as above - is it possible to display these texts as messages in the actual PR conversation rather than requiring the user to click through to the jenkins log?

SchrodingersGat:
the kicad-library repo now has Travis integration:

https://lists.launchpad.net/kicad-lib-committers/msg00509.html

Let's consider the best way to roll this out for *all* the pretty repos!

Note that it will be much easier for the pretty repos as we can determine which files have changed through the git info

SchrodingersGat:
@pointhi thanks for this. I have taken the ideas here and added travis config for each active pretty repo. Currently working through activating travis for each one.

Closing this PR as this has now been implemented.

PR #81, opened 2017-02-13
Klc rearranging

SchrodingersGat:
This PR makes some sweeping changes to the KLC scripts, in preparation for the [new KLC rules](https://github.com/SchrodingersGat/kicad-library/wiki/KLC)

Major points:
- schlib checking no longer has an "--enable-extra" option (all rules are always checked). Too many times we have to instruct users to use this flag.
- A number of rules have been renamed / renumbered
- Some refactoring especially in pcb related rules


SchrodingersGat:
@jkriege2 care to take a look at these? Once these are matched to the new rules we can make the KLC wiki page live.

jkriege2:
Hi!

I'll have a closer look this evening ... if possible, could we merge my PR: https://github.com/KiCad/kicad-library-utils/pull/77 first, because with the renames I fear it would make a huge mess otherwise ... maybe I can also try and include some of your refactorings there already, this should make merging this easier ...

What do you think?
JAN

SchrodingersGat:
If you can make the changes as I suggeted in #77 I will merge that, and then merge master into mine :)

SchrodingersGat:
@jkriege2 if you get a chance can you try and break this? There have been many improvements here but also a lot of changes

SchrodingersGat:
There is one philosophical change that has been made here:

Some of the more ambiguous rules now only generate messages (but do not flag an error). That way, we can check the Travis output and determine if the warning is valid or not.

The new scripts also flag a lot more issues than previously

jkriege2:
Hi!

this is a huge effort ... I looked through the code without testing for now: looks good and no obvious stuff ... I will test the tools over the weekend ... sorry don't ahve time before that!

JAN

jkriege2:
fix conflict?

SchrodingersGat:
Ok, fixed :)

schlib/rules/rule3_1.py
```diff
@@ -7,27 +7,74 @@ class Rule(KLCRule):
...
+            if length < 100:
+                self.violating_pins.append(pin)
+                if not err:
+                    self.error("Incorrect pin length:")
+                err = True
+                self.error(' - pin: {0} ({1}), {2}, length={3}'.format(pin['name'], pin['num'], positionFormater(pin), length))
+                
+            # length not multiple of 50 mils flags a warning
+            if length % 50 != 0:
+                self.warning("- pin: {name} ({num}) - length ({length}mils) is not a multiple of 50mils".format(
```
jkriege2: add whitespace `" - ` for indentiona s above!

schlib/rules/rule3_1.py
```diff
@@ -7,27 +7,74 @@ class Rule(KLCRule):
...
+                    name = pin['name'],
+                    num = pin['num'],
+                    length = length))
+                    
+            # length too long flags a warning
+            if length > 200:
+                self.warning("- pin: {name} ({num}) - length ({length}mils) is too long".format(
```
jkriege2: add whitespace `" - ` for indentiona s above!

jkriege2:
I'm now starting to slowly have a look at your script ... so expect several comments over the weekend ;-) ... Some first comments below:

## `checklib.py`: Welcome message
The welcome message reads `Execute checkrule scripts checking 3.* KLC rules in the libraries` ... maybe it would be better to have something like this: `checkrule.py checks KiCAD 4.x schematic symbols for violations of the KLC rules (>v2.0). You can find the KLCs here: https://github.com/KiCad/kicad-library/wiki/Kicad-Library-Convention.`

## `checklib.py`: Unsorted output
The output is unsorted. Checking e.g. `R` from `device.lib  I get:
```
`D:\KiCAD\review.pretty\kicad-library-utils\schlib>checklib.py -vv d:\KiCAD\kicad-library\library\device.lib -c R
R:
  Violating Rule 3.1
    Pin placement
  Violating Rule 3.10
    Part meta-data is filled out as appropriate
  Violating 3.2
    Symbol visual style
    Pins not located on 100mil grid:
     - pin: ~ (1), @ (0, 150)
     - pin: ~ (2), @ (0, -150)
    Incorrect pin length:
     - pin: ~ (1), @ (0, 150), length=50
     - pin: ~ (2), @ (0, -150), length=50
    missing datasheet, please provide a datasheet link if it isn't a generic component
    Component background filled with TRANSPARENT color, recommended is BACKGROUND color
```
So the error messages are all sorted below the rules, and it's hard to assigne them to a rule ... this SHOULD read:
```
`D:\KiCAD\review.pretty\kicad-library-utils\schlib>checklib.py -vv d:\KiCAD\kicad-library\library\device.lib -c R
R:
  Violating Rule 3.1
    Pin placement
    Pins not located on 100mil grid:
     - pin: ~ (1), @ (0, 150)
     - pin: ~ (2), @ (0, -150)
    Incorrect pin length:
     - pin: ~ (1), @ (0, 150), length=50
     - pin: ~ (2), @ (0, -150), length=50
  Violating Rule 3.10
    Part meta-data is filled out as appropriate
    missing datasheet, please provide a datasheet link if it isn't a generic component
  Violating 3.2
    Symbol visual style
    Component background filled with TRANSPARENT color, recommended is BACKGROUND color
```

## `checklib.py`: representation of rule
If run with `-v` you get e.g. 
```
R:
  Violating Rule 3.1
    Pin placement
  Violating Rule 3.10
    Part meta-data is filled out as appropriate
  Violating 3.2
    Symbol visual style
```
Here the color is the only indication that e.g. `Pin placement` is not the error message, but the rule-description. Would it be possible to format the output e.g. as follows:
```
R:
  Violating Rule 3.1 (Pin placement)
  Violating Rule 3.10 (Part meta-data is filled out as appropriate)
  Violating 3.2 (Symbol visual style)
```
## `checklib.py`:  Rule 3.1 formatting for Aliases
With the current output `self.info("checking alias: {0}".format(alias))` the rule 3.1 always produces error messages, because the main-script says rule X.Y fails if there are ANY messages, even if all check out (test on `LM555N` from device.lib). Therefore I would propose to reformulate the rules as follows:
```python
def check(self):
        """
        Proceeds the checking of the rule.
        The following variables will be accessible after checking:
            * only_datasheet_missing
        """

        self.only_datasheet_missing = False
        invalid_documentation = 0

        #check part itself
        if self.checkDocumentation(self.component.name, self.component.documentation):
            invalid_documentation += 1
        #check all its aliases too
        if self.component.aliases:
            for alias in self.component.aliases.keys():
                #self.info("checking alias: {0}".format(alias))
                if self.checkDocumentation(alias, self.component.aliases[alias], indentation=2, isAlias=True):
                    invalid_documentation+=1
                    print('    ', invalid_documentation)

        return invalid_documentation>0 


    def checkDocumentation(self, name, documentation, indentation=0, isAlias=False):
        prefix=''
        if isAlias: prefix="Alias '{name}': ".format(name=name)
        if not documentation:
            self.error(" "*indentation+prefix+"missing whole documentation (description, keywords, datasheet)")
            return True

        elif (not documentation['description'] or
            not documentation['keywords'] or
            not documentation['datasheet']):

            if (not documentation['description']):
                self.error(" "*indentation+prefix+"missing description")
            if (not documentation['keywords']):
                self.error(" "*indentation+prefix+"missing keywords")
            if (not documentation['datasheet']):
                self.warning(" "*indentation+prefix+"missing datasheet, please provide a datasheet link if it isn't a generic component")
                if (documentation['description'] and
                    documentation['keywords']):
                    self.only_datasheet_missing = True

            # counts as violation if only datasheet is missing and verbosity is high
            if self.verbosity and self.verbosity > Verbosity.NORMAL:
                return True

            return not self.only_datasheet_missing

        elif name.lower() in documentation['description'].lower():
            self.warning( " "*indentation + prefix+"symbol name should not be included in description")
            return True

        return False
```
This way the alias-name is mentioned besides the fail-nessages and no output is generated when all aliases check out. Note the additional bool-parameter `isAlias` of `checkDocumentation()` and the `prefix` that is calculated from it!

Output before my changes:
```
LM555N:
  Violating Rule 3.10
    Part meta-data is filled out as appropriate
    checking alias: NE555
    checking alias: ICM7555
    checking alias: LMC555
    checking alias: TLC555
    checking alias: NA555
    checking alias: SE555
    checking alias: SA555
```

Output after changes:

```
LM555N - No errors
```
Or with a manufactured error (deleted keywords from NE555's DCM entry:
```
LM555N:
  Violating Rule 3.10
    Part meta-data is filled out as appropriate
      Alias 'NE555': missing keywords
```

## Text of error messages
The text for many messages is very short, e.g. `Component line thickness {0}, recommended {1}` (KLC 3.2).
I would suggest to make that into whole senteces, or at least have a verb in there ... that would really improve readability and clearness of the messages (especially for people not beeing acustomed to funny compiler error messages such as in the combination of C++ and templates ;-): `Component line thickness is {0}mil, but recommended thickness is {1}mil`

## `checklib.py`: KLC 3.3
I rewrote some parts of the script:
- better FIX not supported outputs 
- check for visibility should be `pin['pin_type'].startswith('N')`
- indention of listed outputs

```python

# -*- coding: utf-8 -*-

from rules.rule import *

class Rule(KLCRule):
    """
    Create the methods check and fix to use with the kicad lib files.
    """
    def __init__(self, component):
        super(Rule, self).__init__(component, '3.3', 'Invalid pin stacking')
        self.different_names=False
        self.NC_stacked=False
        self.different_types=False
        self.only_one_visible=False

    def stackStr(self, stack):
        multi_unit = int(self.component.definition['unit_count']) > 1
        unit_str = " (unit {u})".format(stack['u']) if multi_unit else ""
        
        # WHY are pins flipped vertically? Mega sad face :(
        return "Pinstack @ ({x},{y}){u}".format(
            x = int(stack['x']),
            y = -1 * int(stack['y']),
            u = unit_str)
        
    def pinStr(self, pin):
        multi_unit = int(self.component.definition['unit_count']) > 1
        return "Pin {name} ({num}){unit}".format(
            name = pin['name'],
            num = pin['num'],
            unit = " (unit {u})".format(u=pin['unit']) if multi_unit else "")
        
    def check(self):
    
        # List of lists of pins that are entirely duplicated
        self.duplicated_pins = []
        
        pin_locations = []
        
        for pin in self.component.pins:
            
            # To be "identical", all following properties much be the same
            
            pinx = pin['posx']    # x coordinate
            piny = pin['posy']    # y coordinate
            pinu = pin['unit']    # unit (for multi-unit parts)
            pinc = pin['convert'] # convert (de morgan)
            
            dupe = False
            
            for loc in pin_locations:
            
                locx = loc['x']
                locy = loc['y']
                locu = loc['u']
                locc = loc['c']
                
                if pinx == locx and piny == locy and pinu == locu and pinc == locc:
                    loc['pins'].append(pin)
                    dupe = True
                    
            if not dupe:
                new_loc = {'x': pinx, 'y': piny, 'u': pinu, 'c': pinc}
                new_loc['pins'] = [pin]
                pin_locations.append(new_loc)
                    
        err = False
                    
        for loc in pin_locations:
            if len(loc['pins']) > 1:
                
                pin_units = set()
                pin_nums = set()
                pin_names = set()
                pin_etypes = set()
                
                vis_pin_count = 0
                
                for pin in loc['pins']:
                    pin_nums.add(pin['num'])
                    pin_names.add(pin['name'])
                    pin_units.add(pin['unit'])
                    pin_etypes.add(pin['electrical_type'])
                    
                    # Add visibile pins
                    if not pin['pin_type'].startswith('N'):
                        vis_pin_count += 1
                    
                    # NC pins should never be stacked
                    if pin['electrical_type'] == 'N':
                        self.error("NC {pin} @ ({x},{y})is stacked on other pins".format(
                            pin = self.pinStr(pin),
                            x = pin['posx'],
                            y = -1*int(pin['posy'])))
                        err = True
                        self.NC_stacked=True
                            
                # Fewer pin numbers than pins
                if len(pin_nums) < len(loc['pins']):
                    self.error("Duplicate pins @ ({x},{y}):".format(
                        x = loc['x'],
                        y = -1 * int(loc['y'])))
                    err = True
                    for pin in loc['pins']: 
                        self.error("  - " + self.pinStr(pin))
                            
                    # If ALL pins are identical, go to next group (no further checks needed)
                    if len(pin_nums) == len(pin_names) == len(pin_units) == 1:
                        self.duplicated_pins.append([pin for pin in loc['pins']])
                        continue
                            
                # Different names!
                if len(pin_names) > 1:
                    self.error(self.stackStr(loc) + " have different names:")
                    err = True
                    for pin in loc['pins']:
                        self.error("  - " + self.pinStr(pin))
                        self.different_names=True
                            
                # Different types!
                if len(pin_etypes) > 1:
                    self.error(self.stackStr(loc) + " have different types:")
                    err = True
                    for pin in loc['pins']:
                        self.error("  - {pin} : {etype}".format(
                            pin = self.pinStr(pin),
                            etype = pinElectricalTypeToStr(pin['electrical_type'])))
                        self.different_types=True
            
                # Only one pin should be visible
                if not vis_pin_count == 1:
                    self.error(self.stackStr(loc) + " must have exactly one (1) invisible pin:")
                    err = True
                    for pin in loc['pins']:
                        self.error("- {pin} is {vis}".format(
                            pin = self.pinStr(pin),
                            vis = 'INVISIBLE' if pin['pin_type'].startswith('N') else 'VISIBLE'))
                        self.only_one_visible=True



        return err
                    
    def fix(self):
    
        # Delete duplicate pins
        if len(self.duplicated_pins) > 0:
            self.info("Removing duplicate pins:")
            
            for pin_groups in self.duplicated_pins:
                # Leave first pin and delete all others
                pin = pin_groups[0]
                
                count = 0
                # Iterate through component pins
                i = 0
                while i < len(self.component.drawOrdered):
                    
                    el = self.component.drawOrdered[i]
                    if not el[0] == 'X': # Pins
                        i += 1 
                        continue
                    
                    p_test = el[1]
                
                    # All these keys must be identical!
                    keys = ['name','num','unit','posx','posy','convert']
                    
                    # Found duplicate
                    if all([p_test[key] == pin[key] for key in keys]):
                        count += 1
                        # Skip the first instance, delete all others
                        if count > 1:
                            del self.component.drawOrdered[i]
                            self.info("Deleting {pin} @ ({x},{y})".format(
                                pin = self.pinStr(pin),
                                x = pin['posx'],
                                y = pin['posy']))
                            continue
                    i += 1
                    
        if self.different_names:
            self.info("FIX for 'different pin names' not supported (yet)! Please fix manually.")
        if self.NC_stacked:
            self.info("FIX for 'NC pins stacked' not supported! Please fix manually.")
        if self.different_types:
            self.info("FIX for 'different pin types' not supported (yet)! Please fix manually.")
        if self.only_one_visible:
            self.info("FIX for 'only one pin in a pin stack is visible' not supported (yet)! Please fix manually.")
```


## `checklib.py`: KLC 3.7
I rewrote some parts of the script:
- better output of current pin-type when that is wrong
- check for visibility should be `visible = not (pin['pin_type'].startswith('N'))`


```python
# -*- coding: utf-8 -*-

from rules.rule import *
import re

class Rule(KLCRule):

    #No-connect pins should be "N"
    NC_PINS = ['^nc$', '^dnc$', '^n\.c\.$']


    #check if a pin name fits within a list of possible pins (using regex testing)
    def test(self, pinName, nameList):

        for name in nameList:
            if re.search(name,pinName,flags=re.IGNORECASE) is not None:
                return True

        return False

    """
    Create the methods check and fix to use with the kicad lib files.
    """
    def __init__(self, component):
        super(Rule, self).__init__(component, '3.7', 'NC pin checking')
        self.etypetxts={'I': 'Input', 'O': 'Outout', 'B': 'Bidirectional', 'T': 'Tristate', 'P': 'Passive', 'C': 'open Collector', 'E': 'open Emitter', 'N': 'Not-connected', 'U': 'Unspecified', 'W': 'power input', 'w':  'power output', '': '???'}
            
    def checkNCPin(self, pin):
        
        err = False
    
        name = pin['name'].lower()
        etype = pin['electrical_type']

        visible = not (pin['pin_type'].startswith('N'))
        
        # Check NC pins
        if self.test(name.lower(), self.NC_PINS) or etype == 'N':
        
            # NC pins should be of type N
            if not etype == 'N': # Not set to NC
                err = True
                self.error("Pin {n} ({name}) should be of type NOT CONNECTED, but is {pintype}".format(
                            n = pin['num'],
                            name = name,
                            pintype=self.etypetxts[etype]))
                                
            # NC pins should be invisible
            if visible:
                err = True
                self.error("Pin {n} ({name}) is NC and VISIBLE. but should be INVISIBLE".format(
                            n = pin['num'],
                            name = name))
                            
        if err:
            self.nc_errors.append(pin)
            
        return err
        
    def check(self):
        """
        Proceeds the checking of the rule.
        The following variables will be accessible after checking:
            * probably_wrong_pin_types
            * double_inverted_pins
        """
        
        self.nc_errors = []
        
        fail = False
        
        for pin in self.component.pins:
                
            if self.checkNCPin(pin):
                fail = True
        
        return fail
        
    def fix(self):
        """
        Proceeds the fixing of the rule, if possible.
        """
        self.info("Fixing...")
        
        for pin in self.nc_errors:
            if not pin['electrical_type'] == 'N':
                pin['electrical_type'] = 'N'
                self.info("Changing pin {n} type to NO_CONNECT".format(n=pin['num']))
            
            if not pin['pin_type'] == 'N':
                pin['pin_type'] = 'N'
                self.info("Setting pin {n} to INVISIBLE".format(n=pin['num']))

        self.recheck()

```

## `checklib.py`: KLC 3.12
I would suggest additional checks:
```python
# -*- coding: utf-8 -*-

from rules.rule import *
import re

class Rule(KLCRule):
    """
    Create the methods check and fix to use with the kicad lib files.
    """
    def __init__(self, component):
        super(Rule, self).__init__(component, 'Rule 3.12', 'Footprint filters')

    def check(self):
        """
        Proceeds the checking of the rule.
        """
        
        filters = self.component.fplist
        
        if len(filters) == 0:
            self.warning("No footprint filters defined")
        else:
            fcnt=1
            for filter in filters:
                if not filter.endswith('*'):
                    self.warning("FPFilter {fcnt}: footprint filter '{filter}' does not end with an asterisk '*'".format(filter=filter, fcnt=fcnt))
                if re.search('(SOIC|SOIJ|SIP|DIP|SO|SOT-\d+|SOT\d+|QFN|DFN|QFP|SOP|TO-\d+|VSO|PGA|BGA|LLC|LGA)-\d+[W-_\*\?$]+', filter, flags=re.IGNORECASE) is not None:
                    self.warning("FPFilter {fcnt}: footprint filter '{filter}' seems to contain pin-number, but should not!".format(filter=filter, fcnt=fcnt))
                if ('-' in filter) or ('_' in filter):
                    self.warning("FPFilter {fcnt}: Minuses and underscores in footprint filter '{filter}' should be escaped with '?' or '*'.".format(filter=filter, fcnt=fcnt))
                fcnt=fcnt+1
        return False

    def fix(self):
        """
        Proceeds the fixing of the rule, if possible.
        """
        self.info("FIX: not supported")
```

## `checklib.py` Rest
The rest of the symbol-check-code seems fine to me.

## Example file
Here's an example lib with errors that I used to test some stuff:
```
EESchema-DOCLIB  Version 2.0
#
$CMP Test_FPFiltersFail
D description
K keywords
F http://link
$ENDCMP
#
$CMP Test_NCnotNC
D description
K keywords
F http://link
$ENDCMP
#
$CMP Test_NCnotNCandVisible
D description
K keywords
F http://link
$ENDCMP
#
$CMP Test_NCstacked
D description
K keywords
F http://link
$ENDCMP
#
$CMP Test_NCvisible
D description
K keywords
F http://link
$ENDCMP
#
$CMP Test_PowerOK
D description
K keywords
F http://link
$ENDCMP
#
$CMP Test_PowerWrongBorder
D description
K keywords
F http://link
$ENDCMP
#
$CMP Test_PowerWrongType
D description
K keywords
F http://link
$ENDCMP
#
$CMP Test_StackError
D description
K keywords
F http://link
$ENDCMP
#
$CMP Test_StackOK
D description
K keywords
F http://link
$ENDCMP
#
$CMP Test_allOK
D description
K keywords
F http://link
$ENDCMP
#
$CMP Test_doubleInvert
D description
K keywords
F http://link
$ENDCMP
#
$CMP Test_duplicatedPin
D description
K keywords
F http://link
$ENDCMP
#
$CMP Test_noFPFilter
D description
K keywords
F http://link
$ENDCMP
#
$CMP Test_noKeywords
D description
F http://link
$ENDCMP
#
$CMP Test_wrongPinAlign
D description
K keywords
F http://link
$ENDCMP
#
$CMP Test_wrongPinLen
D description
K keywords
F http://link
$ENDCMP
#
#End Doc Library
```

```
EESchema-LIBRARY Version 2.3
#encoding utf-8
#
# Test_FPFiltersFail
#
DEF Test_FPFiltersFail U 0 40 Y Y 1 F N
F0 "U" 0 200 50 H V C CNN
F1 "Test_FPFiltersFail" 0 -200 50 H V C CNN
F2 "" 0 200 50 H I C CNN
F3 "" 0 200 50 H I C CNN
$FPLIST
 R*
 R
 SOT-23-5
 SOT-23-5*
 SOIC-8_3.9x4.9mm_Pitch1.27mm
 SOIC-8_3.9x4.9mm_Pitch1.27mm*
 SOIC*3.9x4.9mm*Pitch1.27mm*
 SOIC-16W_7.5x10.3mm_Pitch1.27mm*
 SOIC*7.5x10.3mm*Pitch1.27mm*
$ENDFPLIST
DRAW
S -200 100 200 -150 0 1 10 f
X A 1 -300 0 100 R 50 50 1 1 I
X B 2 300 0 100 L 50 50 1 1 I
X NC 3 -300 -100 100 R 50 50 1 1 N N
ENDDRAW
ENDDEF
#
# Test_NCnotNC
#
DEF Test_NCnotNC U 0 40 Y Y 1 F N
F0 "U" 0 200 50 H V C CNN
F1 "Test_NCnotNC" 0 -200 50 H V C CNN
F2 "" 0 200 50 H I C CNN
F3 "" 0 200 50 H I C CNN
$FPLIST
 R*
$ENDFPLIST
DRAW
S -200 100 200 -150 0 1 10 f
X A 1 -300 0 100 R 50 50 1 1 I
X B 2 300 0 100 L 50 50 1 1 O
X NC 3 -300 -100 100 R 50 50 1 1 B N
ENDDRAW
ENDDEF
#
# Test_NCnotNCandVisible
#
DEF Test_NCnotNCandVisible U 0 40 Y Y 1 F N
F0 "U" 0 200 50 H V C CNN
F1 "Test_NCnotNCandVisible" 0 -200 50 H V C CNN
F2 "" 0 200 50 H I C CNN
F3 "" 0 200 50 H I C CNN
$FPLIST
 R*
$ENDFPLIST
DRAW
S -200 100 200 -150 0 1 10 f
X A 1 -300 0 100 R 50 50 1 1 I
X B 2 300 0 100 L 50 50 1 1 O
X NC 3 -300 -100 100 R 50 50 1 1 T
ENDDRAW
ENDDEF
#
# Test_NCstacked
#
DEF Test_NCstacked U 0 40 Y Y 1 F N
F0 "U" 0 200 50 H V C CNN
F1 "Test_NCstacked" 0 -200 50 H V C CNN
F2 "" 0 200 50 H I C CNN
F3 "" 0 200 50 H I C CNN
$FPLIST
 R*
$ENDFPLIST
DRAW
S -200 100 200 -150 0 1 10 f
X A 1 -300 0 100 R 50 50 1 1 I
X B 2 300 0 100 L 50 50 1 1 I
X NC 3 -300 -100 100 R 50 50 1 1 N N
X NC 4 -300 -100 100 R 50 50 1 1 N N
ENDDRAW
ENDDEF
#
# Test_NCvisible
#
DEF Test_NCvisible U 0 40 Y Y 1 F N
F0 "U" 0 200 50 H V C CNN
F1 "Test_NCvisible" 0 -200 50 H V C CNN
F2 "" 0 200 50 H I C CNN
F3 "" 0 200 50 H I C CNN
$FPLIST
 R*
$ENDFPLIST
DRAW
S -200 100 200 -150 0 1 10 f
X A 1 -300 0 100 R 50 50 1 1 I
X B 2 300 0 100 L 50 50 1 1 I
X NC 3 -300 -100 100 R 50 50 1 1 N
ENDDRAW
ENDDEF
#
# Test_PowerOK
#
DEF Test_PowerOK U 0 40 Y Y 1 F N
F0 "U" -150 150 50 H V C CNN
F1 "Test_PowerOK" 350 -400 50 H V C CNN
F2 "" 0 200 50 H I C CNN
F3 "" 0 200 50 H I C CNN
$FPLIST
 R*
$ENDFPLIST
DRAW
S -200 100 200 -300 0 1 10 f
X A 1 -300 0 100 R 50 50 1 1 I
X B 2 300 0 100 L 50 50 1 1 I
X NC 3 -300 -100 100 R 50 50 1 1 N N
X VCC 4 0 200 100 D 50 50 1 1 W
X GND 5 0 -400 100 U 50 50 1 1 W
ENDDRAW
ENDDEF
#
# Test_PowerWrongBorder
#
DEF Test_PowerWrongBorder U 0 40 Y Y 1 F N
F0 "U" -150 150 50 H V C CNN
F1 "Test_PowerWrongBorder" 500 -350 50 H V C CNN
F2 "" 0 200 50 H I C CNN
F3 "" 0 200 50 H I C CNN
$FPLIST
 R*
$ENDFPLIST
DRAW
S -200 100 200 -300 0 1 10 f
X A 1 -300 0 100 R 50 50 1 1 I
X B 2 300 0 100 L 50 50 1 1 I
X NC 3 -300 -100 100 R 50 50 1 1 N N
X VCC 4 0 200 100 D 50 50 1 1 W
X GND 5 100 200 100 D 50 50 1 1 W
X GND 6 300 -200 100 L 50 50 1 1 W
X VCC 7 -100 -400 100 U 50 50 1 1 W
ENDDRAW
ENDDEF
#
# Test_PowerWrongType
#
DEF Test_PowerWrongType U 0 40 Y Y 1 F N
F0 "U" -150 150 50 H V C CNN
F1 "Test_PowerWrongType" 500 -350 50 H V C CNN
F2 "" 0 200 50 H I C CNN
F3 "" 0 200 50 H I C CNN
$FPLIST
 R*
$ENDFPLIST
DRAW
S -200 100 200 -300 0 1 10 f
X A 1 -300 0 100 R 50 50 1 1 I
X B 2 300 0 100 L 50 50 1 1 I
X NC 3 -300 -100 100 R 50 50 1 1 N N
X VCC 4 0 200 100 D 50 50 1 1 w
X GND 5 0 -400 100 U 50 50 1 1 I
ENDDRAW
ENDDEF
#
# Test_StackError
#
DEF Test_StackError U 0 40 Y Y 1 F N
F0 "U" 0 200 50 H V C CNN
F1 "Test_StackError" 0 -200 50 H V C CNN
F2 "" 0 200 50 H I C CNN
F3 "" 0 200 50 H I C CNN
$FPLIST
 R*
$ENDFPLIST
DRAW
S -200 100 200 -100 0 1 10 f
X A 1 -300 0 100 R 50 50 1 1 I
X B 2 300 0 100 L 50 50 1 1 I
X B 3 -300 0 100 R 50 50 1 1 I N
ENDDRAW
ENDDEF
#
# Test_StackOK
#
DEF Test_StackOK U 0 40 Y Y 1 F N
F0 "U" 0 200 50 H V C CNN
F1 "Test_StackOK" 0 -200 50 H V C CNN
F2 "" 0 200 50 H I C CNN
F3 "" 0 200 50 H I C CNN
$FPLIST
 R*
$ENDFPLIST
DRAW
S -200 100 200 -100 0 1 10 f
X A 1 -300 0 100 R 50 50 1 1 I
X B 2 300 0 100 L 50 50 1 1 I
X A 3 -300 0 100 R 50 50 1 1 I N
ENDDRAW
ENDDEF
#
# Test_allOK
#
DEF Test_allOK U 0 40 Y Y 1 F N
F0 "U" 0 200 50 H V C CNN
F1 "Test_allOK" 0 -200 50 H V C CNN
F2 "" 0 200 50 H I C CNN
F3 "" 0 200 50 H I C CNN
$FPLIST
 R*
$ENDFPLIST
DRAW
S -200 100 200 -150 0 1 10 f
X A 1 -300 0 100 R 50 50 1 1 I
X B 2 300 0 100 L 50 50 1 1 I
X NC 3 -300 -100 100 R 50 50 1 1 N N
ENDDRAW
ENDDEF
#
# Test_doubleInvert
#
DEF Test_doubleInvert U 0 40 Y Y 1 F N
F0 "U" 0 200 50 H V C CNN
F1 "Test_doubleInvert" 0 -200 50 H V C CNN
F2 "" 0 200 50 H I C CNN
F3 "" 0 200 50 H I C CNN
$FPLIST
 R*
$ENDFPLIST
DRAW
S -200 100 200 -150 0 1 10 f
X A 1 -300 0 100 R 50 50 1 1 I
X ~B 2 300 0 100 L 50 50 1 1 I I
X NC 3 -300 -100 100 R 50 50 1 1 N N
ENDDRAW
ENDDEF
#
# Test_duplicatedPin
#
DEF Test_duplicatedPin U 0 40 Y Y 1 F N
F0 "U" 0 200 50 H V C CNN
F1 "Test_duplicatedPin" 0 -200 50 H V C CNN
F2 "" 0 200 50 H I C CNN
F3 "" 0 200 50 H I C CNN
$FPLIST
 R*
$ENDFPLIST
DRAW
S -200 100 200 -150 0 1 10 f
X A 1 -300 0 100 R 50 50 1 1 I
X B 2 300 0 100 L 50 50 1 1 B
X B2 2 300 -100 100 L 50 50 1 1 O
X NC 3 -300 -100 100 R 50 50 1 1 N N
ENDDRAW
ENDDEF
#
# Test_noFPFilter
#
DEF Test_noFPFilter U 0 40 Y Y 1 F N
F0 "U" 0 200 50 H V C CNN
F1 "Test_noFPFilter" 0 -200 50 H V C CNN
F2 "" 0 200 50 H I C CNN
F3 "" 0 200 50 H I C CNN
DRAW
S -200 100 200 -100 0 1 10 f
X A 1 -300 0 100 R 50 50 1 1 I
X B 2 300 0 100 L 50 50 1 1 I
ENDDRAW
ENDDEF
#
# Test_noKeywords
#
DEF Test_noKeywords U 0 40 Y Y 1 F N
F0 "U" 0 200 50 H V C CNN
F1 "Test_noKeywords" 0 -200 50 H V C CNN
F2 "" 0 200 50 H I C CNN
F3 "" 0 200 50 H I C CNN
$FPLIST
 R*
$ENDFPLIST
DRAW
S -200 100 200 -100 0 1 10 f
X A 1 -300 0 100 R 50 50 1 1 I
X B 2 300 0 100 L 50 50 1 1 I
ENDDRAW
ENDDEF
#
# Test_wrongPinAlign
#
DEF Test_wrongPinAlign U 0 40 Y Y 1 F N
F0 "U" 0 150 50 H V C CNN
F1 "Test_wrongPinAlign" 0 -250 50 H V C CNN
F2 "" 0 150 50 H I C CNN
F3 "" 0 150 50 H I C CNN
$FPLIST
 R*
$ENDFPLIST
DRAW
S -200 50 200 -150 0 1 10 f
X A 1 -300 -50 100 R 50 50 1 1 I
X B 2 300 -50 100 L 50 50 1 1 I
ENDDRAW
ENDDEF
#
# Test_wrongPinLen
#
DEF Test_wrongPinLen U 0 40 Y Y 1 F N
F0 "U" 0 200 50 H V C CNN
F1 "Test_wrongPinLen" 0 -200 50 H V C CNN
F2 "" 0 200 50 H I C CNN
F3 "" 0 200 50 H I C CNN
$FPLIST
 R*
$ENDFPLIST
DRAW
S -200 100 200 -150 0 1 10 f
P 2 0 1 10 250 100 200 100 N
X A 1 -300 0 100 R 50 50 1 1 I
X B125 2 300 0 125 L 50 50 1 1 I
X NC 3 -300 -100 100 R 50 50 1 1 N N
X C150 4 350 -100 150 L 50 50 1 1 I
X C150 5 400 100 150 L 50 50 1 1 I
ENDDRAW
ENDDEF
#
#End Library

```

pcb/rules/klc_constants.py
```diff
@@ -0,0 +1,23 @@
+
+# silkscreen layer
+KLC_SILK_WIDTH = 0.12
+KLC_SILK_WIDTH_ALLOWED = [0.10, 0.12, 0.15]
```
jkriege2: whould we really "allow" these? Why not just check for 0.12mm ... otherwise we will never get a consistent lib (I think)

pcb/check_kicad_mod.py
```diff
@@ -65,32 +76,47 @@
...
-
-                # example of customized printing feedback by checking the rule name
-                # and a specific variable of the rule
-                #if rule.name == 'Rule 6.6' and len(rule.f_courtyard_all) == 0:
-                #    printer.red('No courtyard line found in the module', indentation=4)
+                
         if args.fix:
             rule.fix()
             if args.verbose:
                 if len(rule.fix_verbose_message)>0:
```
jkriege2: this causes an error: use `fix_message` instead of `fix_verbose_message`

pcb/check_kicad_mod.py
```diff
@@ -65,32 +76,47 @@
...
                     vm=rule.fix_verbose_message.split('\n');
```
jkriege2: should be `vm=rule.fix_message;` as `fix_message` is now a list and not a string!

pcb/rules/rule.py
```diff
@@ -4,16 +4,27 @@ class KLCRule(object):
...
     """
-    def __init__(self, module, args, name, description,verbose_message='',fix_verbose_message=''):
+    def __init__(self, module, args, name, description,verbose_message='',fix_message=''):
         self.module = module
         self.name = name
         self.description = description
-        self.verbose_message=verbose_message
-        self.fix_verbose_message=fix_verbose_message
-        self.args=args
+
```
jkriege2: missing `self.args=args` this is required by e.g. rule 7.5 to check whether `--fixmore` was given!

pcb/rules/rule7_4.py
```diff
@@ -0,0 +1,159 @@
...
+                graph['width'] = KLC_FAB_WIDTH
+                
+        if self.checkMissingValue():
+            module.value['layer'] = 'F.Fab'
+            module.value['font']['height'] = self.expected_val_width
+            module.value['font']['width'] = self.expected_val_width
+            module.value['font']['thickness'] = self.expected_val_thickness
+            
+        if self.checkSecondRef():
+            pass
```
jkriege2: Why did you delete the creation of a second %R-label? Code was present in `pcb/rules/rule10_5.py`

```python
if module.needs_secondrefdes:
                txt={'user': '%R', 'pos': module.secondrefdespos, 'layer': module.secondrefdeslayer, 'font': module.secondrefdesfont, 'hide': False } 
                module.userText.append(txt)
                print('added second REFDES')
            else:
                for txt in range(0,len(module.userText)):
                    if module.userText[txt]['user']=='%R':
                        print('OLD: ', module.userText[txt])
                        module.userText[txt]['pos']=module.secondrefdespos
                        module.userText[txt]['layer']=module.secondrefdeslayer
                        module.userText[txt]['font']=module.secondrefdesfont
                        print('NEW: ', module.userText[txt])
                        print('fixed second REFDES ', txt)
                        break

```

pcb/rules/rule8_2.py
```diff
@@ -9,7 +9,7 @@ class Rule(KLCRule):
     Create the methods check and fix to use with the kicad_mod files.
     """
     def __init__(self, module, args):
-        super(Rule, self).__init__(module, args, 'Rule 6.4', 'For surface-mount devices, footprint anchor is placed in the middle with respect to device lead ends. (IPC-7351).')
+        super(Rule, self).__init__(module, args, 'Rule 8.2', 'For surface-mount devices, footprint anchor is placed in the middle of the footprint (IPC-7351).')
 
     def check(self):
         """
```
jkriege2: I think there should be a message output for the violation, e.g. add `self.addMessage("Achor not placed placed in the middle of the footprint (IPC-7351).")` before `return True` in line 35

jkriege2: also add fix-message in `fix()`, e.g.: `self.addFixMessage("Anchor fixed.")`

pcb/rules/rule8_1.py
```diff
@@ -0,0 +1,40 @@
...
+        return False
+
+
+    def fix(self):
+        """
+        Proceeds the fixing of the rule, if possible.
+        """
+        module = self.module
+        if self.check():
+            module.attribute = 'smd'
```
jkriege2: add message `self.addFixMessage("Set Surface Mount attribute.")`

pcb/rules/rule8_3.py
```diff
@@ -0,0 +1,65 @@
...
+        
+        return any([self.checkPad(pad) for pad in module.filterPads('smd')])
+        
+    def fix(self):
+        """
+        Proceeds the fixing of the rule, if possible.
+        """
+        module = self.module
+        
+        for pad in module.filterPads('smd'):
```
jkriege2: add fix-message, e.g.:

```python
        anychanged=False
        for pad in module.filterPads('smd'):
            if pad['layers'] != self.required_layers: anychanged=True
            pad['layers'] = self.required_layers

        if anychanged: self.addFixMessage("Layers on some SMD-pads have been fixed.")
```

pcb/rules/rule9_1.py
```diff
@@ -0,0 +1,40 @@
...
+        return False
+
+
+    def fix(self):
+        """
+        Proceeds the fixing of the rule, if possible.
+        """
+        module = self.module
+        if self.check():
+            module.attribute = 'pth'
```
jkriege2: add fix message `self.addFixMessage("Set 'Through Hole' placement type.")`

pcb/rules/rule9_2.py
```diff
@@ -33,7 +33,7 @@ def check(self):
                 if x == 0 and y == 0:
                     anchor_ok = True
                 else:
-                    self.verbose_message=self.verbose_message+"pin1 was found at x={0}mm y={1}mm. ".format(x,y)
+                    self.addMessage("pin1 was found at x={0}mm y={1}mm. ".format(x,y))
 
             return not anchor_ok
```
jkriege2: add fix message `self.addFixMessage("Moved Anchor to pin 1.")`

pcb/rules/rule9_4.py
```diff
@@ -0,0 +1,65 @@
...
+        module = self.module
+        
+        return any([self.checkPad(pad) for pad in module.filterPads('thru_hole')])
+        
+    def fix(self):
+        """
+        Proceeds the fixing of the rule, if possible.
+        """
+        module = self.module
+
```
jkriege2: add fix message 

```python
        anychanged=False
        for pad in module.filterPads('thru_hole'):
            if pad['layers'] != self.required_layers: anychanged=True
            pad['layers'] = self.required_layers

        if anychanged: self.addFixMessage("Layers on some through-hole pads have been fixed.")

```

pcb/rules/rule9_5.py
```diff
@@ -0,0 +1,48 @@
...
+            * pin1_count
+        """
+        module = self.module
+        
+        return any([self.checkPad(pad) for pad in module.filterPads('thru_hole')])
+        
+    def fix(self):
+        """
+        Proceeds the fixing of the rule, if possible.
+        """
```
jkriege2: add no fix supported message

pcb/rules/rule9_6.py
```diff
@@ -0,0 +1,75 @@
...
+        """
+        module = self.module
+        
+        return any([self.checkPad(pad) for pad in module.filterPads('thru_hole')])
+        
+    def fix(self):
+        """
+        Proceeds the fixing of the rule, if possible.
+        """
+        pass
```
jkriege2: add no fix supported message

jkriege2:
... and fix conflicts ;-)

Hope my comments above help ...

best and GREAT WORK ON THIS!
JAN

pcb/rules/klc_constants.py
SchrodingersGat: IPC7351C allows for any of those three values. I think we should be permissive here, otherwise we'll never be able to properly conform. Default is still 0.12mm

SchrodingersGat:
Jan I think I have hit all the points you have raised. LMK if there's anything further you can see. Thanks for the review.

pcb/rules/klc_constants.py
jkriege2: Ahh ... just saw this is in the KLC now! ... then fine with me!

SchrodingersGat:
I have reworked all the kicad_mod rules so they now work in the same way as the lib rules.

Instead of addMessage() you can now call error() warning() and info() 

Lots more fixes too.

I have also tried to consolidate all the rules such that the output is the same (or similar) for each rule.

SchrodingersGat:
I have now reworked the courtyard rule as I believe it was too complicated. It checks the following:

1. If there are no courtyard elements present, error
1. If any courtyard elements are incorrect width, error
1. If any courtyard elements are incorrect grid, error

It no longer calculates the actual clearance as the checks were very convoluted and would not work for courtyards that use circles / arc and have complex shapes! This will be best done as a human check I think.

The --fix option now adds a conservative rectangular courtyard if (and only if) there are currently NO graphical elements on the courtyard layers.

Anything else should be done by hand.

I have also included some new helper functions and written a BoundingBox class which makes all this much easier. It also means that if we decide to make these rules more complicated in the future it will be easier to do so.

SchrodingersGat:
@jkriege2 I think this is working really well now - can you have another look and let me know if you spot any further errors?

SchrodingersGat:
There is now a fix option to add second Refdes to Fab layer

jkriege2:
I will have a look later ... about the courtyard-check: I would really like to keep the warning at least ... I think with the improvements I did in https://github.com/KiCad/kicad-library-utils/pull/88 it worked rather nicely for rectangular courtyards (still the majority of courtyards! I think this warning really helps, especially since checking the courtyards by hand is very painfull! Why not keep my proposed way, which is basically yours, just added that existing rectangular courtyards are fixed if (and only if) `--fixmore` is given ... A warning is issued anyways ... Maybe I can incorporate my code in your version and post a proposal here (like I did above for other rules) ... let's see ...

JAN

jkriege2:
Hi!

I'm just testing your tools ... some things I already found:

## `%R`-Position
For THT there is a weird positioning:
2017-03-08 19_07_03-footprint editor no active library
I would suggest to put `%R` for THT initially to the same location as the REFDES on SilkS... That's IMHO the best an automatism can achieve.

## Output doubled
Also There is doubled-output when using with `--fix` ... is that by design?
```
D:\KiCAD\review.pretty\kicad-library-utils\pcb>check_kicad_mod.py -vv d:\KiCAD\Housings_DIP.pretty\DIP-4_W10.16mm.kicad_mod --fix
Checking footprint 'DIP-4_W10.16mm':
  Violating Rule 7.4
    Fabrication layer requirements
    Second Reference Designator missing
     - Add RefDes to F.Fab layer with '%R'
  Violating Rule 7.4
    Fabrication layer requirements
    Second Reference Designator missing
     - Add RefDes to F.Fab layer with '%R'
    Adding second RefDes to F.Fab layer @ 3.4925,1.27
Please, resave the files using KiCad to keep indentation standard.
```

Best,
JAN

jkriege2:
auto-generated courtyard is too large:
2017-03-08 19_16_41-footprint editor no active library

In this case it SHOULD be 0.25mm, but is way larger at left/right, but OK top/bottom!

pcb/kicad_mod.py
```diff
@@ -2,25 +2,50 @@
...
-            y[key]=value
-    return y
+from boundingbox import BoundingBox
+
+
+# Rotate a point by given angle (in degrees)
+def _rotatePoint(point, degrees):
+
+    # Create a new point (copy)
+    p = {}
```
jkriege2: this should be p=point

and the for-loop can go away!

jkriege2: PS: The for-loop lead to an exception for me ...

schlib/rules/rule3_12.py
```diff
@@ -0,0 +1,51 @@
...
+    
+        self.bad_filters = []
+   
+        for filter in filters:
+            errors = []
+            # Filter must contain a "*" wildcard
+            if not '*' in filter:
+                errors.append("Does not contain wildcard ('*') character")
+                
+            if len(filter.count(':')) > 1:
```
jkriege2: this crashes ... should be simply `filter.count(':') > 1`

jkriege2:
Also the output of the schlib-checker is doubled when running fix():

```
`D:\KiCAD\review.pretty\kicad-library-utils\schlib>checklib.py -vv test.lib --fix
Checking symbol 'Test_FPFiltersFail':
  Violating Rule 3.12 - Footprint filters
    Footprint filters should match all appropriate footprints
    Footprint filter 'R' not correctly formatted
     - Does not contain wildcard ('*') character
    Footprint filter 'SOT-23-5' not correctly formatted
     - Does not contain wildcard ('*') character
    Footprint filter 'SOIC-8_3.9x4.9mm_Pitch1.27mm' not correctly formatted
     - Does not contain wildcard ('*') character
  Violating Rule 3.12 - Footprint filters
    Footprint filters should match all appropriate footprints
    Footprint filter 'R' not correctly formatted
     - Does not contain wildcard ('*') character
    Footprint filter 'SOT-23-5' not correctly formatted
     - Does not contain wildcard ('*') character
    Footprint filter 'SOIC-8_3.9x4.9mm_Pitch1.27mm' not correctly formatted
     - Does not contain wildcard ('*') character
    FIX: not supported
Checking symbol 'Test_NCnotNC':
  Violating Rule 3.7 - Unused pins
    Unused pins should be set as NOT CONNECTED and should be INVISIBLE
    NC pins are not correct pin-type:
     - Pin NC (3) @ (-300,-100) should be of type NOT CONNECTED, but is of type BIDI
  Violating Rule 3.7 - Unused pins
    Unused pins should be set as NOT CONNECTED and should be INVISIBLE
    NC pins are not correct pin-type:
     - Pin NC (3) @ (-300,-100) should be of type NOT CONNECTED, but is of type BIDI
    Fixing...
    Changing pin 3 type to NO_CONNECT
    Everything fixed
```

schlib/rules/rule3_12.py
```diff
@@ -0,0 +1,51 @@
...
+
+    def checkFilters(self, filters):
+    
+        self.bad_filters = []
+   
+        for filter in filters:
+            errors = []
+            # Filter must contain a "*" wildcard
+            if not '*' in filter:
+                errors.append("Does not contain wildcard ('*') character")
```
jkriege2: This should really check for an '*' at the end!

jkriege2: I think this text will not really help anewby ... better (also minding the comment above):
`This footprint filter should end with an "*".`

schlib/rules/rule3_12.py
```diff
@@ -0,0 +1,51 @@
...
+                
+            if len(filter.count(':')) > 1:
+                errors.append("Filter should not contain more than one (':') character")
+
```
jkriege2: What about my proposals:
```
                if re.search('(SOIC|SOIJ|SIP|DIP|SO|SOT-\d+|SOT\d+|QFN|DFN|QFP|SOP|TO-\d+|VSO|PGA|BGA|LLC|LGA)-\d+[W-_\*\?$]+', filter, flags=re.IGNORECASE) is not None:
                    self.warning("FPFilter {fcnt}: footprint filter '{filter}' seems to contain pin-number, but should not!".format(filter=filter, fcnt=fcnt))
                if ('-' in filter) or ('_' in filter):
                    self.warning("FPFilter {fcnt}: Minuses and underscores in footprint filter '{filter}' should be escaped with '?' or '*'.".format(filter=filter, fcnt=fcnt))

```

These are designed to help and note errors that are often done when people submit a symbol!

jkriege2:
otherwise at least the schlib-checker looks good ...

Best,
JAN

SchrodingersGat:
I have fixed the bug with incorrect '%R' positioning for THT parts.

Also fixed the double message output.

I have added your footprint filter warnings too. 

Did I miss anything?

SchrodingersGat:
I have now added a check for invalid pad types, as per https://github.com/KiCad/kicad-library/issues/1096

cedricboudinet:
Hello
I am running the scripts under Debian. There are some errors because linux path separator is '\', so sys.path.append("..\common") should be replaced by sys.path.append(os.path.join('..',"common")).
Bests

jkriege2:
Hi!

I'm currently having another look at this: schlib seems fine to me ... didn't spot anything that stoof off.

Some things remain with the kicad_mod:
1. autoplacing `%R` in the center is OK for many components, but not for all (especially those with some drawing and/or text in the center), see e.g. 
2017-03-11 17_19_53-footprint editor no active library
2017-03-11 17_22_52-footprint editor no active library
I still kind of have the feeling that placing `%R` at the same spot as REFDES is the better way to go, as the user usually already placed that appropriately ... but I really like the scaled text when e.g. fixing an 0805 or 1206 resistor for `%R` 👍 
2. default courtyard looks fine to me!
3. I would still like to have the check of existing rectangular courtyards ... I will try and add that when we have merged this PR ...
4. I don't see anything other immediate, apart from the point @cedricboudinet made above!

Best & hope we can merge this soon,
JAN

PS: If I find anything more, I will let you know ;-)

SchrodingersGat:
@cedricboudinet thanks, I have updated the scripts for better path management

SchrodingersGat:
@jkriege2 responses below:

1. I think this is OK - we are not suggesting that the --fix option is pushed straight to the libraries unchecked. It is a good "starting point" and there are going to be cases where it doesn't work and needs to be tweaked. But for the **general case** we want the second RefDes to be on the F.Fab layer at the middle of the part.
2. Great!
3. Yes, I still do like that option, it's currently causing a lot of noise but we can make it work I am sure
4. Fixed

jkriege2:
OK, so should I merge?

cedricboudinet:
@SchrodingersGat : one sys.path.append("..\..\common") still remains in pcb/rules/rule7_5.py

Regards

pcb/kicad_mod.py
```diff
@@ -627,33 +692,47 @@ def filterArcs(self, layer):
...
+            
+            r = math.sqrt(dx*dx + dy*dy)
+            
+            bb.addPoint(cx, cy, radius=r)
+            
+        # Add all arcs
+        arcs=self.filterArcs(layer)
         for c in arcs:
-            r=math.sqrt((c['start']['x']-c['end']['x'])*(c['start']['x']-c['end']['x'])+(c['start']['y']-c['end']['y'])*(c['start']['y']-c['end']['y']))
+            cx = c['center']['x']
```
jkriege2: this needs to be `cx = c['start']['x']`

pcb/kicad_mod.py
```diff
@@ -627,33 +692,47 @@ def filterArcs(self, layer):
...
+            cy = c['center']['y']
```
jkriege2: this needs to be `cx = c['start']['y']`

pcb/rules/rule9_2.py
```diff
@@ -45,4 +47,5 @@ def fix(self):
         """
         module = self.module
         if self.check() and len(self.pin1_position)>0:
+            self.info("Moved anchor position to Pin-1")
             module.setAnchor(min(self.pin1_position))
```
jkriege2: remove the min: `module.setAnchor(self.pin1_position)`

pcb/rules/rule9_2.py
```diff
@@ -20,22 +20,24 @@ def check(self):
...
-            self.pin1_count = len(pads)
-            self.pin1_position = []
-
-            # check/get pads positions
-            anchor_ok = False
+                
+            if len(pads) == 0:
+                self.error("Pad 1 not found in footprint")
+                return True
+
```
jkriege2: The two varoables `self.pin1_count` and `self.pin1_position` are no longer set: 

```python
self.pin1_count=len(pads)
            if len(pads) == 0:
                self.error("Pad 1 not found in footprint")
                return True
                
            for pad in pads:
                pos = pad['pos']
                if len(self.pin1_position)==0: 
                    self.pin1_position=[pos['x'],pos['y']]
```

2017-03-12 17_53_45-pcb_rules_rule9_2 py_ working tree - tortoisegitmerge

jkriege2:
One more thing: Shouldn't we have a check for KLC 10.4.ii that suggests adding `${KISYS3DMOD}/` if it's not there (just warning, no error though).

SchrodingersGat:
All done, I think :)

jkriege2:
Maybe you can provide a 3D model link ... so the model is found, when available in future?

Otherwise fine and I chan merge!

JAN

SchrodingersGat:
@jkriege2 I am not sure what you mean

jkriege2:
Ooopps ... that was misplaced ... I think I wanted to post that for one of the other reviews I did yesterday ... Please ignore ;-)

SchrodingersGat:
I have done some work improving the footprint file reading and writing:

* Fixed a lot of parsing errors
* Improved output formatting
* the --fix option should now keep original file formatting, only showing diffs

pcb/kicad_mod.py
```diff
@@ -47,29 +69,29 @@ def __init__(self, filename):
...
-        self.layer = self._getValue('layer')
+        self.layer = self._getValue('layer', 'pth', 2)
 
         # locked flag
-        self.locked = True if self._hasValue(self.sexpr_data, 'locked') else False
+        self.locked = self._getValue('locked', False, 2)
 
         # description
-        self.description = self._getValue('descr')
+        self.description = self._getValue('descr', 2)
```
jkriege2: you omitted the second argument ... if not provided we read the int 2 instead of an empty string:
```python
self.description = self._getValue('descr', '', 2)
```

pcb/kicad_mod.py
```diff
@@ -47,29 +69,29 @@ def __init__(self, filename):
...
 
         # tags
-        self.tags = self._getValue('tags')
-
+        self.tags = self._getValue('tags', 2)
```
jkriege2: you omitted the second argument ... if not provided we read the int 2 instead of an empty string:
```python
self.tags = self._getValue('tags', '', 2)
```

jkriege2:
... and one suggestion. I think the warning "no 3D model given" should be an error and the script should simply add the default model. For that you need to add a function
```python
def addModel(self, filename, pos=[0,0,0], scale=[1,1,1], rotate=[0,0,0]):
        model_dict = {'file':filename}
        # position
        model_dict['pos'] = {'x':pos[0], 'y':pos[1], 'z':pos[2]}
        # scale
        model_dict['scale'] = {'x':scale[0], 'y':scale[1], 'z':scale[2]}
        # rotate
        model_dict['rotate'] = {'x':rotate[0], 'y':rotate[1], 'z':rotate[2]}
        self.models.append(model_dict)
```
to the class in kicad_mod.py

... and modify rule 10.4:
```python
# -*- coding: utf-8 -*-

from rules.rule import *
import os

SYSMOD_PREFIX = "${KISYS3DMOD}/"

class Rule(KLCRule):
    """
    Create the methods check and fix to use with the kicad_mod files.
    """
    def __init__(self, module, args):
        super(Rule, self).__init__(module, args, 'Rule 10.4', '3D model reference')
        self.no3DModel=False

    def checkModel(self, model):
    
        error = False
        
        # Allowed model types
        extensions = ["step","stp","wrl"]
        
        model = model['file']
    
        if model.startswith(SYSMOD_PREFIX):
            model = model.replace(SYSMOD_PREFIX,"")
        else:
            self.warning("Model path should start with '" + SYSMOD_PREFIX + "'")
            
        try:
            model_dir = os.path.dirname(model)
            fn = os.path.basename(model).split(".")
            model_file = ".".join(fn[:-1])
            model_ext = fn[-1]
        except:
            self.error("Model '{mod}' is invalid path".format(mod=model))
            return True
            
        if not model_ext.lower() in extensions:
            self.error("Model '{mod}' is incompatible format (must be STEP or WRL file)".format(mod=model))
            return True
        
        fp_dir = self.module_dir[0] + ".3dshapes"
        fp_name = self.module.name
        
        if not model_dir == fp_dir:
            self.error("3D model directory is different from footprint directory (found '{n1}', should be '{n2}')".format(n1=model_dir, n2=fp_dir))
            error = True
            
        if not model_file == fp_name:
            # Exception for footprints that have additions e.g. "_ThermalPad"
            if fp_name.startswith(model_file):
                self.warning("3D model name is different from footprint name (found '{n1}', should be '{n2}')".format(n1=model_file, n2=fp_name))
            else:
                self.error("3D model name is different from footprint name (found '{n1}', should be '{n2}')".format(n1=model_file, n2=fp_name))
                error = True
            
        if not isValidName(model_file):
            error = True
            self.error("3D model path '{p}' contains invalid characters as per KLC 1.7".format(
                p = model_file))
            
        return error
        
    def check(self):
        """
        Proceeds the checking of the rule.
        The following variables will be accessible after checking:
            * module_dir
            * model_dir
            * model_file
        """
        module = self.module
        module_dir = os.path.split(os.path.dirname(os.path.realpath(module.filename)))[-1]
        self.module_dir = os.path.splitext(module_dir)

        models = module.models
        
        if len(models) == 0:
            # Warning msg
            self.error("No 3D model provided")
            self.no3DModel=True
            return True
        
        if len(models) > 1:
            self.warning("More than one 3D model provided")
            
        model_error = False
            
        for model in models:
            if self.checkModel(model):
                model_error = True
                
        return model_error

    def fix(self):
        if self.no3DModel:
            fp_dir = self.module_dir[0] + ".3dshapes/"
            fp_name = self.module.name
            model=SYSMOD_PREFIX+fp_dir+fp_name+".wrl"
            self.module.addModel(model)
            self.info("added default 3D model link '"+model+"'")
        else:
            self.info("Fix not supported")

```

What do you think?

jkriege2:
PS: I checked the `--fix`-option here: https://github.com/KiCad/Crystals.pretty/pull/22 and using the two fixes/additions above it worked out nicely :-)))

I think we should move towards merging?

jkriege2:
... I also think that the warning "Model path should start with '${KISYS3DMOD}/'" should be fixed automatically ... maybe it would be good to react to these with a switch `--fixwarnings`? .. the the check()-methods would need one of 3 return values: ERROR/WARNING/OK ...

What do you think?

pcb/kicad_mod.py
```diff
@@ -701,143 +635,350 @@ def filterPads(self, pad_type):
         for pad in self.pads:
             if pad['type'] == pad_type:
                 pads.append(pad)
+                
+        pads = sorted(pads, key = lambda p : str(p['number']))
 
         return pads
+        
+    # Get the middle position between pads
+    def padMiddlePosition(self, pads=None):
```
jkriege2: This function should be reformulated to:

```python
def padMiddlePosition(self, pads=None):
    bb = self.padsBounds(pads)
    return bb.center
```

Otherwise this function calculates the center of mass of the center-positions of each pad. If then the pads are distributed unevenly between left and right (e.g. see https://github.com/KiCad/Pin_Headers.pretty/blob/master/Pin_Header_Straight_1x39_Pitch2.54mm_SMD_Pin1Left.kicad_mod ) the center pin is asked to be ofset a bit (0.05mm to the left or right from the (IMHO correct) center-line

SchrodingersGat:
@jkriege2 I've fixed the bugs, thanks for that. I think I will leave the 3D model checks as they are for now, let's focus on getting this merged first and then we can improve with new ideas.

SchrodingersGat:
@jkriege2 I have now tested these scripts on Linux with both Python2 and Python3 (they needed some work to make them work across both Python version as I was using some Python3 features).

Anything else that stands out to you?

SchrodingersGat:
@poeschlr @jkriege2 any final comments on this? :)

poeschlr:
Sorry i have seen this pull request now for the first time. (I forgot to add the library utils to my watch list. Yes you mentioned something over at the klc discussion issue but i didn't connect the dots.) 
Hopefully i will find time do a few tests tomorrow. (It's nearly midnight here right now.)

SchrodingersGat:
Thanks :)

jkriege2:
no more comments from me ... let me know when I should merge!

JAN

poeschlr:
I played around a bit. (used my connectors as test subjects and manipulated the footprints such that specific klc rules are violated.)

Until now i could only find one minor problem. If i move my tht connectors such that they are centered, both rule 7.5 and 9.2 are violated. When using the fix extension, the courtyard is not on grid afterwards. (I think the script corrects the courtyard before moving the part such that pin 1 is at the correct position. If i run the script a second time everything is correct afterwards.)

(I found loads of problems in my own footprints while testing the scripts but this is another matter. Expect some new pull requests shortly.)

SchrodingersGat:
@poeschlr it sounds like the issue you have found should not affect the Travis CI as it does not use the --fix option anywhere. We should definitely look into this but I'd like to get this merged first. 

@jkriege2 can you please merge this as KLC2.0 is [now live!](https://forum.kicad.info/t/klc-2-0-now-with-bells-and-whistles/5985)

jkriege2:
Great Work!

Although I'm a bit low on time currently (and will be off for 2 weeks of holidays starting on saturday) ... I will sonn start to work on getting the courtyard checks back online ...

Best,
JAN

PR #85, opened 2017-02-20
Implemented code to find pins with identical coordinate

jophut:
The code first makes a list sorted by primary position x, secondary position y, tertiary pin number.
Then it compares subsequent items for equal coordinates.
If items with equal coordinates are found it stores them in a new list.
This way, coordinates containing 2 (or more) pins are grouped.

Finally the group is printed per coordinate. (with increasing pin number because of the sorting order)

Not sure if the code should be in 'rule3_1.py', but for now a good starting point.


SchrodingersGat:
Hi @jophut this is a great addition, thanks for the contribution.

We are almost finished with the new KLC (v2.0) - see https://github.com/SchrodingersGat/kicad-library/wiki/KLC

I have copied your code and it will feature in the new KLC. I'm closing this PR unmerged with the current KLC, hope you don't mind.

PR #87, opened 2017-02-26
Improvements to checklib

SchrodingersGat:
- Extra checks only generate warnings (not errors)
- Ignores changes in line endings
- Fixed a logic bug where error would be flagged inappropriately

@jkriege2 @pointhi this PR is to improve the Travis output. It should now ignore line-ending changes and also only generate warnings for extra-checking

### DIFF
--- a/schlib/schlib.py
+++ b/schlib/schlib.py
@@ -36,7 +36,7 @@ class Documentation(object):
 
         else:
             if not os.path.isfile(self.filename):
-                sys.stderr.write("DCM file '{filename}' does not exist".format(filename=self.filename))
+                sys.stderr.write("DCM file '{filename}' does not exist\n".format(filename=self.filename))
                 return
             else:
                 self.validFile = True
--- a/schlib/schlib.py
+++ b/schlib/schlib.py
@@ -36,7 +36,7 @@ class Documentation(object):
 
         else:
             if not os.path.isfile(self.filename):
-                sys.stderr.write('"{filename}" does not exist'.format(filename=self.filename))
+                sys.stderr.write("DCM file '{filename}' does not exist".format(filename=self.filename))
                 return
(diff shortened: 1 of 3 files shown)


SchrodingersGat:
I have also reduced the amount of "noise" the comparelibs script generates

SchrodingersGat:
@jkriege2 can you please quickly check this one so we can fix the Travis issues? :

jkriege2:
Hi!

I can check it this evening ...

Best,
JAN

schlib/schlib.py
```diff
@@ -36,7 +36,7 @@ def __init__(self, filename, create = False):
 
         else:
             if not os.path.isfile(self.filename):
-                sys.stderr.write('"{filename}" is not a valid .dcm file\n'.format(filename=self.filename))
+                sys.stderr.write('"{filename}" does not exist'.format(filename=self.filename))
```
jkriege2: please add a linebreak `\n` at the end of this error message, as currently the output is e.g.:

```
"d:\KiCAD\kicad-library\library\Altera.dcm" does not existchecking component: EP1210
  Violating Rule 3.8
    Description and keywords properties contains information about the component.
```

jkriege2: Also: Maybe a better text:
```
  'The DCM-file for library "{filename_lib}" does not exist (expected name: "{filename}").'
```

Basically that's clear, but for newbiew its easier to talk about the .lib, as the editor mostly hides the DCM from them.

jkriege2:
PS: Otherwise it's fine and works for the cases I checked.

best,
JAN

SchrodingersGat:
Should be good to go now @jkriege2

jkriege2:
OK!

nice work  ;-)

JAN

PR #88, opened 2017-02-27
Bugfixes for Rule 6.6 (CourtYard)

jkriege2:
- better claculation of actual courtyard clearance 
- now uses offsets in the pad center

### DIFF
--- a/pcb/kicad_mod.py
+++ b/pcb/kicad_mod.py
@@ -726,7 +726,8 @@ class KicadMod(object):
             offset={'x':0.0, 'y':0.0}
             if 'drill' in pad:
                 if 'offset' in pad['drill']:
-                    offset=pad['drill']['offset']
+                    if 'x' in pad['drill']['offset']:
+                        offset=pad['drill']['offset']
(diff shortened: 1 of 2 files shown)


jkriege2:
@SchrodingersGat if you have time: This should reduce the noise by courtyard-checks a bit!

SchrodingersGat:
I am close to having the new KLC ready to go - check updates here - https://github.com/KiCad/kicad-library-utils/pull/81

Can we hold off on this PR until the new KLC is in? At this stage it's going to just create merge errors for the new KLC

jkriege2:
hmmm ... we can, but then the rule won't work correctly in some cases and give even more strange outputs ... but you can merge your's first and the I try to merge mine in?

CarlPoirier:
Hi @jkriege2 @SchrodingersGat,

Is this PR still relevant? If so, I can review it once the conflicts are resolved.

SchrodingersGat:
I believe that @jkriege2 was going to look into this once more now that KLC 2.0 is out.

jkriege2:
I think we can close this for now ... and I will work on an updated soluation, based on the current status of the scripts ;-)

Issue #96, opened 2017-04-01
kicad_mod.py fails to save file after issue fixing.


4x5dm:
Error log from console:
```D:\MyProjects\kicad-library-utils\pcb>python check_kicad_mod.py -v --fixmore d:\MyProjects\kicad_libraries\custom_kicad_footprints.pretty\Connector_Mini-DIN_Female_6Pin_2rows.kicad_mod
To print using colors you have to install colorama. Try to install it using: "pip install colorama"
[Continuing using no color mode]

checking module: Connector_Mini-DIN_Female_6Pin_2rows
  Violating Rule 6.6
    Courtyard line has a width 0.05mm. This line is placed so that its clearance is measured from its
    center to the edges of pads and body, and its position is rounded on a grid of 0.01mm.
      A courtyard clearance in the range 0.25...1.0mm was found, but 0.5mm was expected.
      The recommendet courtyard rectangle would be:
          {'y': -5.05, 'x': -9.81, 'height': 14.05, 'width': 17.0}mm.
        REPLACING Courtyard with RECTANGLE with clearance 0.5mm on layer F.CrtYd
Traceback (most recent call last):
  File "check_kicad_mod.py", line 132, in <module>
    module.save()
  File "D:\MyProjects\kicad-library-utils\pcb\kicad_mod.py", line 829, in save
    f = open(filename, 'w', newline='\r\n')
TypeError: 'newline' is an invalid keyword argument for this function
```

I can fix this issue and create a pull request.

4x5dm:
I use Python 2.7.

SchrodingersGat:
Hi @4x5dm please download the scripts again - there have been some major changes since yesterday. 

This issue should be fixed so let me know how you go. Thanks.

4x5dm:
All works. Thanks!

PR #97, opened 2017-04-01
Fixed file save fail

4x5dm:
Fixed file save fail shown in issue #96 .


poeschlr:
The open function of python 3 is not compatible to the one used in python 2.x
The solution is to use io.open (This function has the same signature as open in python 3)

See:
http://stackoverflow.com/questions/40491158/newline-does-not-work-with-python-2-7

https://docs.python.org/3/howto/pyporting.html#text-versus-binary-data

SchrodingersGat:
Hi @4x5dm thanks for raising this, but the brand new KLC scripts (see https://github.com/KiCad/kicad-library-utils/pull/81) are about to be merged and they do not use this keyword, so the problem should be fixed).

SchrodingersGat:
@4x5dm can you confirm if the problem still exists in the new scripts?

4x5dm:
All works. Thanks!

Issue #104, opened 2017-04-05
Incorrect check for Vss pin in rule4_5.py


hackscribble:
Latest KLC rule 4.5.ii treats Vss as a negative power pin which, like ground pins, should be at the bottom of the symbol.  The module `rule4_5.py` treats Vss as a positive power pin.

```
    def checkPowerPins(self):
        PWR = ['^[ad]*v(aa|cc|dd|ss|bat|in)$']
    
        first = True
        
        for pin in self.component.pins:
            name = str(pin['name'].lower())
            for pwr in PWR:
                if re.search(pwr, name, flags=re.IGNORECASE) is not None:
                    # Pin orientation should be "down"
                    if not pin['direction'] == 'D':
                        if first:
                            first = False
                            self.warning("Power pins should be placed at top of symbol")
                        self.warningExtra(pinString(pin))
```

I can work on a PR to move Vss to `checkGroundPins()` if that helps?

**EDIT**

Also, the script does not check for "V+" or "V-" which are mentioned in 4.5.i and 4.5.ii.

SchrodingersGat:
@hackscribble good find, I'll look into this.

`V+` and `V-` are not included in this check because often they are input pins e.g. for differential amps etc. Adding these to this check raises too many false positives.

hackscribble:
I have updated the PR to remove V+ and V-.

SchrodingersGat:
Closed, merged

PR #106, opened 2017-04-05
Fix handling of negative power pins in Rule 4.5 check

hackscribble:
See https://github.com/KiCad/kicad-library-utils/issues/104

Moved Vss from positive power pin check function to ground pin check function. Added V+ to positive check and V- to ground check. Updated warning messages.


hackscribble:
Updated PR to remove V+ and V-.  See: https://github.com/KiCad/kicad-library-utils/issues/104#issuecomment-292390575

SchrodingersGat:
Thanks for the fix @hackscribble

Issue #98, opened 2017-04-03
Rule 8.2 violation with asymmetric SMD package


hackscribble:
I'm designing a footprint for a TO-252-5-11 package (see images below).  The footprint fails check_kicad_mod.py with a rule 6.4 violation.  

Using the --fix option and comparing the resulting file with my original, it looks like the program calculates footprint centre based on the position of the **centres** of the pads.  For this type of package (the tab pad is much wider than the pin pads), this doesn't seem to correctly interpret the KLC rule, assuming that one should treat the tab as a lead:

>  "Generally [the middle] is the centroid calculated with respect to the **device lead ends**"

 

Am I interpreting this correctly, and therefore can ignore this error?

Thanks.

#### Package

In X direction (red axis) distance from end of pins to end of tab is 9.98mm.  Distance from left edge of pin pads to right edge of tab pad is 10.3mm.  So I assume that package sits centred within pads with 0.3mm clearance on each side.

![to-252-5-11 package](https://cloud.githubusercontent.com/assets/7076027/24604729/e91adec8-185d-11e7-88f3-d887fd047020.png)

#### Draft footprint

![to-252-5-11](https://cloud.githubusercontent.com/assets/7076027/24604747/fe825afc-185d-11e7-8274-dfbdb64931c4.png)

hackscribble:
Attaching my original footprint and the version after --fix.

[footprints.zip](https://github.com/KiCad/kicad-library-utils/files/889692/footprints.zip)

SchrodingersGat:
Hi @hackscribble please refer to the [brand new KLC 2.0](https://github.com/KiCad/kicad-library/wiki/Kicad-Library-Convention) - this rule is now **8.2**

In this case the center of the footprint should be set to the datasheet position not the center position of the pads, as the exception to 8.2 states :)

hackscribble:
Thanks @SchrodingersGat.

The datasheet does not specify an origin for pick and place, which leaves ...

> Footprint anchor should be placed in the middle of the footprint (IPC-7351). Generally this is the centroid calculated with respect to the device lead ends.

So I guess I am OK to ignore the rule violation error, since I have calculated the centre based on lead ends.

Is there a new version of check_kicad_mod.py in the pipeline which will have the new KLC numbering?

SchrodingersGat:
> Is there a new version of check_kicad_mod.py in the pipeline which will have the new KLC numbering?

Yes, it should be merged any day now :)

[...]

PR #107, opened 2017-04-05
Rule 8.2: changed rule check to use overpadBounds() instead of padBounds()

hackscribble:
See https://github.com/KiCad/kicad-library-utils/issues/98

### Change

```
    # Get the middle position between pads
    # Use the outer dimensions of pads to handle footprints with pads of different sizes
    def padMiddlePosition(self, pads=None):
        
        bb = self.overpadsBounds(pads)
        return bb.center
```

### Test data

[footprints.zip](https://github.com/KiCad/kicad-library-utils/files/900628/footprints.zip)

| Footprint | All pads same size | Anchor at centre|
|-|-|-|
| SOT-23| Yes | Yes |
| SOT-23_moved| Yes | No |
| TO-252-5-11| No | Yes |
| TO-252-2Lead| No | No |

### `check_kicad_mod.py` result for Rule 8.2

| Footprint | Expected result | Result with master | Result with PR |
|-|-|-|-|
| SOT-23|Pass| Pass | Pass |
| SOT-23_moved|Fail| Fail |Fail |
| TO-252-5-11|Pass| Fail <sup>1</sup> | Pass |
| TO-252-2Lead|Fail| Pass <sup>2</sup> | Fail |

__Notes__

<sup>1</sup> This incorrect result is what led to the issue being raised.

<sup>2</sup> This "false positive" for an existing footprint may be due to the footprint having been fixed before release by using the current version of `check_kicad_mod.py` with the `--fix` argument.

### DIFF
--- a/pcb/kicad_mod.py
+++ b/pcb/kicad_mod.py
@@ -652,9 +652,10 @@ class KicadMod(object):
         return pads
         
     # Get the middle position between pads
+    # Use the outer dimensions of pads to handle footprints with pads of different sizes
     def padMiddlePosition(self, pads=None):
         
-        bb = self.padsBounds(pads)
+        bb = self.overpadsBounds(pads)
         return bb.center
 
     def padsBounds(self, pads=None):


SchrodingersGat:
Thanks for the contribution @hackscribble

PR #108, opened 2017-04-07
Added check for non-empty DATASHEET field

SchrodingersGat:
Also added a check for extra (non-standard) fields

### DIFF
--- a/schlib/rules/rule4_9.py
+++ b/schlib/rules/rule4_9.py
@@ -16,7 +16,7 @@ class Rule(KLCRule):
     def checkEmpty(self, field):
         if 'name' in field.keys():
             name = field['name']
-            if name:
+            if name and name not in ['""', "''"] and len(name) > 0:
                 return False
         return True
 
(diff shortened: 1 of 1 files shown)


schlib/rules/rule4_9.py
```diff
@@ -8,106 +8,127 @@ class Rule(KLCRule):
...
     def __init__(self, component):
         super(Rule, self).__init__(component, 'Rule 4.9 - Default fields', 'Default fields contain the correct information')
-        
+
     def checkVisibility(self, field):
         return field['visibility'] == 'V'
-        
+
+    # return True if a field is empty else false
+    def checkEmpty(self, field):
```
jkriege2: I don't really understand that part of the code (this function) ... what does it do?
At least for me the scripts in this PR always report KLC-violation 4.9 :
```
Checking symbol 'Pololu_Breakout_DRV8825':
  Violating Rule 4.9 - Default fields
    Default fields contain the correct information
    Datasheet field must be EMPTY
```

SchrodingersGat:
My bad, I had updated that line but didn't commit it properly. Should be fixed now.

jkriege2:
yep, everything fine!

nice work
JAN

Issue #103, opened 2017-04-04
Unstaged files


hackscribble:
Just cloned the repo and found two unstaged files in master branch (see below).  

If I fork and create a feature branch to submit a PR, can I safely commit these two files, or do they need to be tidied up first?

Thanks.

```
ray@ray-Inspiron-7537:~/kicad-library-utils$ git status
On branch master
Your branch is up-to-date with 'origin/master'.
Changes not staged for commit:
  (use "git add <file>..." to update what will be committed)
  (use "git checkout -- <file>..." to discard changes in working directory)

	modified:   schlib/autogen/R_NETWORKS/make_R_NET.py
	modified:   schlib/autogen/make_DIPSwitches/make_sw_dip.py

no changes added to commit (use "git add" and/or "git commit -a")
```

hackscribble:
I have submitted PRs without committing these files.  Hope that's OK.

PR #114, opened 2017-04-10
Committing two unstaged files.

hackscribble:
See https://github.com/KiCad/kicad-library-utils/issues/103

Not sure what is meant to happen with these two files.  Suggesting a commit as a way of tidying up master branch, but maybe they should be deleted?


poeschlr:
Maybe line endings?
(Because it seems all lines differ but nothing is shown in the github diff.)

PR #115, opened 2017-04-11
Changed different 3D model name from error to warning

SchrodingersGat:
The "different 3D model" error generates too many false positives.

Changed to a warning so that it is still visible in the Travis logs but does not fail the build


SchrodingersGat:
@jkriege2 any thoughts on this? :)

jkriege2:
fine with me!
JAN

PR #116, opened 2017-05-06
exception for KLC 4.3 which allows stacks with one poer input/output and all other pins PASSIVE and invisible

jkriege2:
This is useful, especially for voltage regulators where several pins are tied together on the output ... to not cause ERC-violations, only one of the stacked pins may be POWER_OUT/IN all others should be PASSIVE (and invisible).

This fix enforces that!

Best,
JAN

### DIFF
--- a/schlib/rules/rule4_3.py
+++ b/schlib/rules/rule4_3.py
@@ -38,6 +38,7 @@ class Rule(KLCRule):
         return pinString(pin, unit)
         
     def check(self):
+        self.component.padInSpecialPowerStack=set();
     
         # List of lists of pins that are entirely duplicated
         self.duplicated_pins = []
@@ -150,6 +151,7 @@ class Rule(KLCRule):
                         # in special pin stacks the power-input/power-output/output pin has to be visible and the passive pins need to be invisible
                         specialpincount=0
                         for pin in loc['pins']:
+                            self.component.padInSpecialPowerStack.add(pin['num'])
                             # check if all passive pins are invisible
                             if pin['electrical_type']=='P' and (not pin['pin_type'].startswith('N')):
                                 self.errorExtra("{pin} : {etype} should be invisible (power-pin stack)".format(
@@ -186,7 +188,7 @@ class Rule(KLCRule):
                             pin = self.pinStr(pin),
                             vis = 'INVISIBLE' if pin['pin_type'].startswith('N') else 'VISIBLE'))
                         self.only_one_visible=True
-						
+                        
         # check for invisible power I/O-pins (unless in power.lib)
         isPowerLib=(self.component.reference=='#PWR')
         if (not err) and (not isPowerLib):
--- a/schlib/rules/rule4_3.py
+++ b/schlib/rules/rule4_3.py
@@ -181,7 +181,7 @@ class Rule(KLCRule):
                         
                 # Only one pin should be visible (checks have already been done, when isSpecialXPassivePinStack=true)
                 if (not isSpecialXPassivePinStack) and (not vis_pin_count == 1):
-                    self.error(self.stackStr(loc) + " must have exactly one (1) invisible pin")
+                    self.error(self.stackStr(loc) + " must have exactly one (1) visible pin")
                     err = True
                     for pin in loc['pins']:
                         self.errorExtra("{pin} is {vis}".format(
--- a/schlib/rules/rule4_3.py
+++ b/schlib/rules/rule4_3.py
@@ -121,15 +121,34 @@ class Rule(KLCRule):
                         self.different_names=True
                             
                 # Different types!
-                if len(pin_etypes) > 1:
-                    self.error(self.stackStr(loc) + " have different types")
-                    err = True
-                    for pin in loc['pins']:
-                        self.errorExtra("{pin} : {etype}".format(
-                            pin = self.pinStr(pin),
-                            etype = pinElectricalTypeToStr(pin['electrical_type'])))
-                        self.different_types=True
-            
+                if (len(pin_etypes) > 1):
+                    powerpads=False;
+                    if (len(pin_etypes)==2) and ("W" in pin_etypes) and ("P" in pin_etypes):
+                        powerpads=True;
+                    if (len(pin_etypes)==2) and ("w" in pin_etypes) and ("P" in pin_etypes):
+                        powerpads=True;
+                    if not powerpads:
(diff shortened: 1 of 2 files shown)


SchrodingersGat:
Jan this is a great idea but I think there may be some issues with the logic:

A) what if there are more than 2 pads in a stack?
B) I think that multiple power inputs can be stacked without issue (no ERC errors)
C) output pins should not be stacked either as this will trigger an ERC warning too.

jkriege2:
Hi @SchrodingersGat 
ad A) Note that `pin_etypes` is a set, which assembles simply a list of types in the stack: For any normal stack that should be 1 (only one type) ... so that's not an issue, of did I misunderstand your comment?

ad B,C) will fix

JAN

poeschlr:
Are you sure stacking multiple power input pins is ok? Aren't invisible power input pins always global labels? Or is there an exception for stacked pins?

SchrodingersGat:
@poeschlr you are correct - I performed this test:

2017-05-07 17_48_32-eeschema kicad-test _ c__users_oliver_kicad-master

All the invisible pins were joined in PCBNEW.

Thus, power pins cannot be invisible under any case.

I think that if all the invisible pins are *passive* then that is "ok".

This is quickly becoming very messy, though.

SchrodingersGat:
@jkriege2 POWER_INPUT and POWER_OUTPUT pins cannot be made invisible at all.

jkriege2:
OK ... KiCAD definitely does many with with too many non-explicit rules ;-)

Should we keep the rule for OUTPUT-pins?

BTW: I stumbled upon this, when reworking the regul.lib ... so I'll have to put a bit more work into that ;-)

cu
JAN

SchrodingersGat:
What are the cases where stacked pins are currently used extensively? Perhaps we should review whether they are allowed at all. If they are considered useful then we will need to make the rules more explicit.

jkriege2:
I wanted to use them for regul.lib, e.g. it would tidy up an LM317 in an SO8-package to this:
2017-05-07 10_42_09-part library editor d__kicad_test_addlin_ _kicad-library_library_regul lib
instead of this:
2017-05-07 10_45_24-part library editor d__kicad_test_addlin_ _kicad-library_library_regul lib


... but I'm reverting now ;-)

jkriege2:
So, should we close this or at least have the exception for OUTPUT pins?

cu
JAN

SchrodingersGat:
If there is one visible power pin and multiple invisible passive pins, that should be ok, right?

jkriege2:
OK, I think I was confused ... basically the rule was meant to check exactly that:
Now it (should) work. a apsecial-stack is identified if the stack has only two types of pins of which one is passive ath the other W, w or O

In that case, there has to be exactly ONE W,w or O-pin, which needs to be visible and all other pins have to be invisible PASSIVE pins.

OK like that?

Also I added fix()-support!

jkriege2:
... + better comments ;-)

SchrodingersGat:
Better, but I have found a case that it does not check:

If you have a stack with multiple OUTPUT / POWER_INPUT / POWER_OUTPUT pins, this does not give an error! (It should for each case).

jkriege2:
I will check this evening ...

Best,
JAN

jkriege2:
OK, should be fixed now (works for me) + I added a check for invisible power-pins (unless in a power-symbol)

cu
JAN

jkriege2:
Hi @SchrodingersGat !

while working on regul.lib (see https://github.com/KiCad/kicad-library/pull/948) I noticed another problem with the rule changes that I proposed: When KLC 4.3 enforces power-stacks with one power-input/output + invisible PASSIVE pins, subsequently KLC 4.6 will complain about wrong type of power-pins. To overcome this I communicate all pins that are in a special power-stack (and should therefore be excluded from KLC 4.6) via self.component to subsequent rules. Now KLC 4.6 excludes checks on those pins. With these changes I don't get any errors (only warnings, mostly POWER-pins on left/right, as these are voltage regulators) in regul.lib (see PR above).

Could you please have another look at my PR here and there? What do you think? Is this workable for power-pin stacks?

Basically the enforced rule is now: All power-pin-stacks need one visible power-in/out and N>0 invisible PASSIVE pins for the other pins in the stack. This way no power-pin is visible and for power-out we don't get ERC-errors. OK?

cu
JAN

SchrodingersGat:
I think that if you run KLC 4.6 on its own (with `-r`) then it will break because `self.component.padInSpecialPowerStack` is undefined...

SchrodingersGat:
I think that your idea is a good one, though. We'll need to make this obvious in the KLC descriptions.

jkriege2:
Hi!

I proposed an extension to KLC: https://github.com/KiCad/kicad-library/issues/1243

Best,
JAN

jkriege2:
PS: I will look into the `-r` problem later this week,

JAN

jkriege2:
Hi!
tokk me a bit longer, but I fixed the `-r` problem: If 4.6 is checked and outputs an error, but rule 4.3 has not been run (i.e. hasattr(self.component, 'padInSpecialPowerStack') is False) the rule outpurs a message to the user, telling him to also run rule 4.6 as the error might be ignorable:

```
Checking symbol 'MC78L05_SO8':
  Violating Rule 4.6 - Pin types
    Pin electrical type should match pin function
    Power pins should be of type POWER INPUT or POWER OUTPUT
     - NOTE: If power-pins have been stacked, you may ignore this error in some
cases. Ensure to also check rule 4.3 in this case!
     - Pin GND (2) @ (0,-300) is of type PASSIVE
     - Pin GND (3) @ (0,-300) is of type PASSIVE
     - Pin GND (6) @ (0,-300) is of type PASSIVE
```

So running with `-r 4.3,4.6` works:
```
Checking symbol 'MC78L05_SO8' - No errors
```

Can we merge like that?  And also extend the KLC, as proposed there (https://github.com/KiCad/kicad-library/issues/1243)?
Then we could go on with my rework of regul.lib (would be great to also get that merged soon, as there are already several other PRs to that lib + it would allow us to close several other PRs!

Best,
JAN

SchrodingersGat:
@jkriege2 some more thoughts:

Output
`Pinstack @ (x,y) must have exactly one (1) invisible pin`

Please change the word "invisible" to "visible" here

Other than that, it seems great :)

jkriege2:
@SchrodingersGat DONE!

SchrodingersGat:
Thanks Jan for your hard work

PR #118, opened 2017-05-21
Fix for KLC9.6

SchrodingersGat:
Round before comparison to prevent floating point errors


SchrodingersGat:
This is a small fix to prevent the sort of errors seen here:

https://travis-ci.org/KiCad/Connectors_USB.pretty/builds/228486343?utm_source=github_status&utm_medium=notification

PR #121, opened 2017-05-29
Fix for KLC9.6 with oval pads

Ratfink:
While cca77e6 fixed the rounding error for circular holes, the problem still existed for oval-shaped holes.  This commit applies the same method (rounding the annular ring before comparing to the minimum size) to the non-circular pad code path.


SchrodingersGat:
Thanks!

PR #123, opened 2017-06-04
10.4 fix

SchrodingersGat:
\r characters were causing the dictionary lookup for cmp documentation to fail.

This patch trims out these characters


SchrodingersGat:
Ugh, ignore this. It's pushed the wrong thing.

PR #128, opened 2017-06-08
Fixed parameter -p --path (Not working in initial Clone)

PCB4u:
The -p and --path cmdline args are not working in initial clone.
I added os.chdir(base_dir) to go to the path defined with -p or --path cmdline args before execute the "git clone" command.


SchrodingersGat:
Thanks :)

PR #132, opened 2017-06-23
Clarify the --rotate argument direction

evanshultz:
It's not in the help, and for first-time users (or infrequent users) having the rotation direction would be useful.


poeschlr:
@SchrodingersGat or @jkriege2  could one of you confirm that the help text is correct?

evanshultz:
Thanks @poeschlr. Suppose I could have included a screenshot of the help before:
image

And after:
image

We could do a little bit of text adjustment if you still want it to fit all the text on the same line. Suggestions are welcome.

jkriege2:
THanks!
JAN

PR #136, opened 2017-07-09
Added rule check for unix-style line endings

SchrodingersGat:
* Added rule 1.8 for unix-style line endings
* Fixed a bug relating to reading file line endings


jkriege2:
Hi!

I have two concerns:
1. I get a bug on windows ;-)
```
D:\KiCAD\review.pretty\kicad-library-utils\pcb>check_kicad_mod.py -vv R_win.kicad_mod
could not parse module: R_win.kicad_mod
Error: ord() expected string of length 1, but int found
```
2. Shouldn't GIT handle file-endings via `*.kicad_mod text=auto` in `.gitattributes`? If we use this feature, there will be errors for every kicad_mod on windows, as GIT automatically converts line-endings on windows to win-setting and back-converts when submitting ... This feature would only make sense within TRAVIS (maybe we should flag such checks and only use them, when `--travis` is given?)

Best,
JAN

SchrodingersGat:
@jkriege2 hmm that's why your original code didn't work on Linux :) 

We can make this a linux-only test and run it just on Travis, that should ensure that it only catches true line ending errors.

jkriege2:
yes ... the Linux+Travis-only solution would be great ... how can we distinguish?

What about the ord()-function? Did that change between python versions? I ran the script with python 3.4.4 on win7 yesterday (I think).

JAN

SchrodingersGat:
@jkriege2 the test only runs when it detects it is on a linux platform. So this should be hidden for windows users and show automatically on Travis.

jkriege2:
Hi!

the Linux-test looks good ... although I can only check the windows half (will do later today). Did you check it on Linux?

Best,
JAN

PS: please also fix the ord() problem!

SchrodingersGat:
Yep, developed on Linux, it works ok. I have now made it work for Py3 and Py2 (with a dirty hack)

jkriege2:
OK, this is not triggered on Win ... and I trust you thoroughly checked in Linux. So please have a look at my last suggestion (see review comments) ... and after that I merge. OK?

JAN

SchrodingersGat:
@jkriege2 I can't see any review comments?

jkriege2:
Hmmm ... I extended the old review comments ... anyways,
here's my comment:

Code: 
```
 # Ord is different for python3
        if sys.version_info.major == 3:
```

Comment:
maybe >=3 to be prepared for future versions?

Best,
JAN

SchrodingersGat:
Good point, fixed

PR #117, opened 2017-05-15
Add script to check 3D model coverage

hackscribble:
I offer for consideration this script `check_3d_coverage.py` as a candidate to be added to the utilities.  It is intended as a data cleanup / planning tool to help decide where work is required to improve the 3D model collection.  It does **not** replace the detailed validation of the Rule 10.4 check script.

The script builds a list of 3D models referenced in the footprint modules in either (a) all `.pretty` folders on disk, or (b) a specified folder.  It highlights modules without a valid reference to a `.wrl`, `.step` or `.stp` model.

The script then looks for the 3D models in the associated `.3dshapes` folder in the library on disk.  It highlights missing models, and also any `.wrl` models that are not referenced by any footprints.

**EDIT** I have added a command line argument to switch the checking between `kicad-library/modules/packages3d` and the `packages3D` repo.

Sample output ...

```
Checking Potentiometers (contains 128 modules)
- No model file specified in /home/ray/KiCad Contributing/Potentiometers.pretty/Potentiometer_Trimmer-EVM3E.kicad_mod
- No model file specified in /home/ray/KiCad Contributing/Potentiometers.pretty/Potentiometer_Trimmer-EVM3S_R.kicad_mod
- No model file specified in /home/ray/KiCad Contributing/Potentiometers.pretty/Potentiometer_Trimmer_VishaySpectrol_Econtrim-Type36M.kicad_mod
- No model file specified in /home/ray/KiCad Contributing/Potentiometers.pretty/Potentiometer_Trimmer_VishaySpectrol_Econtrim-Type36P.kicad_mod
- No model file specified in /home/ray/KiCad Contributing/Potentiometers.pretty/Potentiometer_Trimmer_VishaySpectrol_Econtrim-Type36T.kicad_mod
- No model file specified in /home/ray/KiCad Contributing/Potentiometers.pretty/Potentiometer_Trimmer_VishaySpectrol_Econtrim-Type36X.kicad_mod
- No model file specified in /home/ray/KiCad Contributing/Potentiometers.pretty/Potentiometer_Trimmer_VishaySpectrol_Econtrim-Type36k.kicad_mod
- No model file specified in /home/ray/KiCad Contributing/Potentiometers.pretty/Potentiometer_WirePads.kicad_mod
- No model file specified in /home/ray/KiCad Contributing/Potentiometers.pretty/Potentiometer_WirePads_Small.kicad_mod
[...]

PR #138, opened 2017-07-24
Added script to check for footprint errors

SchrodingersGat:
Similar to the [3D model script](https://github.com/KiCad/kicad-library-utils/pull/117) submitted by @hackscribble this script checks a directory of .lib files to see where there may be footprint violations

- Checks empty footprint fields
- Checks for invalid footprint formatting
- Checks for footprints that do not exist

A good way to quickly show where effort needs to be directed to fix footprint associations.

### DIFF
--- a/check_symbol_footprints.py
+++ b/check_symbol_footprints.py
@@ -1,0 +1,157 @@
+#!/usr/bin/env python
+# -*- coding: utf-8 -*-
+
+"""
+Why even is this script?
+
+This script takes a bunch of symbol libraries (.lib files),
+and a whole heap of footprint libraries (.pretty directories),
+and determines which symbols need their footprint associations updated.
+
+Tests tested:
+
+a) Symbol is missing a footprint (this is OK, some are generic)
+b) Symbol footprint pattern is illegal (should be of the form <FpLib:FpName>)
+c) Referenced footprint library or footprint name is missing
+
+How to use?
+
+python check_symbol_footprints.py -h
+
+"""
+
+from __future__ import print_function
+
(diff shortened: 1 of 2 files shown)


hackscribble:
Looks good :smile: 

Would it be useful to add a list of the footprint filters (if any), especially for symbols with an empty footprint field?

jkriege2:
Hi!

nice script ... I think to make it even more useful, could you change these things:
1. an example in the `--help`-message would be great, I don't really understand, what I can specify ... e.g. can I also specify a single .lib, or only a directory of libs? Ca I specify a directory of .pretty's or only a single .pretty? ...
2. also: Can wildcards be used?
3. The output is a bit unformatted ... the general error messages, which precede the detailed lists should be highlghted more (e.g. using `=================================================` and/or larger spaces in between ...)
4. Could we also check whether FPFilters are given and whether any of the FPFilters matches the given footprint (if available)? ... maybe this functionality could also be part of the normal lib-check-scripts ?

Best,
JAN

SchrodingersGat:
Note to self : Add facility to search for all symbols that reference a particular footprint

poeschlr:
@SchrodingersGat i think this has been forgotten and now it needs a merge.

poeschlr:
@jkriege2 is this functionality already in our scripts?

Ratfink:
`checklib.py` performs these checks as part of S5.1, if the path to the footprints is given (otherwise, it still checks that the format is right and warns if it thinks there should be a default footprint set).  Closing this as it would be redundant.

Issue #129, opened 2017-06-11
checklib.py verbose flag not implicit when checking component


PhirePhly:
When trying to use schlib/checklib.py to check a single component, it fails with a Traceback. Adding the -v verbose flag causes it to behave as expected and give me the DRC errors. This is confusing since the help text for checklib states: `-c COMPONENT, --component COMPONENT check only a specific component (implicitly verbose)`

Example output without and with -v flag:
```
kenneth@thor:~/src/kicad-library/library$ ../../kicad-library-utils/schlib/checklib.py -c ATMEGA168A-AU ./atmel.lib 
Traceback (most recent call last):
  File "../../kicad-library-utils/schlib/checklib.py", line 68, in <module>
    if (args.verbose>2):
TypeError: unorderable types: NoneType() > int()

kenneth@thor:~/src/kicad-library/library$ ../../kicad-library-utils/schlib/checklib.py -c ATMEGA168A-AU -v ./atmel.lib 
Checking symbol 'ATMEGA168A-AU':
  Violating Rule 4.1 - Pin placement
    Pins must be placed on regular grid, and be of correct length
    Pins not located on 100mil grid:
     - Pin ADC7 (22), @ (-900, -350)
     - Pin (PCINT8/ADC0)PC0 (23), @ (1000, 250)
     - Pin (PCINT9/ADC1)PC1 (24), @ (1000, 150)
     - Pin (PCINT10/ADC2)PC2 (25), @ (1000, 50)
     - Pin (PCINT11/ADC3)PC3 (26), @ (1000, -50)
     - Pin (PCINT12/SDA/ADC4)PC4 (27), @ (1000, -150)
     - Pin (PCINT13/SCL/ADC5)PC5 (28), @ (1000, -250)
     - Pin ADC6 (19), @ (-900, -250)
     - Pin (PCINT14/~RESET~)PC6 (29), @ (1000, -350)
  Violating Rule 4.5 - Pin orientation
    Wherever possible, pins should be arranged by function
    Ground and negative power pins should be placed at bottom of symbol
    Positive power pins should be placed at top of symbol
  Violating Rule 4.6 - Pin types
    Pin electrical type should match pin function
    Pin types should match pin function
  Violating Rule 4.8 - Field text size
    Field text uses a common size of 50mils.
     - Pin (PCINT19/OC2B/INT1)PD3 (1), text size 40, number size 40
     - Pin (PCINT20/XCK/T0)PD4 (2), text size 40, number size 40
[...]

PR #140, opened 2017-07-26
bugfix: crash, when not giving --verbose-option 

jkriege2:
this fixes the reported bug at #129 by @PhirePhly


SchrodingersGat:
Thanks Jan for this fix

PR #141, opened 2017-08-02
[WIP] STM32 autogen update

lorem-ipsum:
I am currently working on an update for the STM32 parts autogenerator. The changes added so far allow the script to run successfully on the current database of chips. The other changes adjust to the evolved KLC.

I've also changed the datasheet links as all datasheets can be accessed via `http://www.st.com/resource/en/datasheet/CHIPNUMBER.pdf`. This also allowed to remove the manual downloading of the pdfs.

Open issues:
- Packages are wrong. Currently the script just takes the name from the chip (e.g. LQFP64). This does not follow KLC and sadly isn't even unique. I am currently working on some hackery to get that information from the datasheets.
- Some chips violate rule 4.5 of KLC as the VBAT pins are not placed correctly. According to checklib they should be placed on top of symbol as they are positive power pins. It just seems to be a warning, should I change it anyways?

### DIFF
--- a/schlib/autogen/stm32/README.md
+++ b/schlib/autogen/stm32/README.md
@@ -5,11 +5,8 @@ This script turns XML files defining pins for STM32 devices into KiCAD libraries
 * XML files, taken from STM32CubeMX install (from db/mcu folder)
 * XML files for obsolete devices (currently STM32F050xx and STM32F313xx), which
   can be taken from STM32CubeMX install (archives in olddb folder)
-* Datasheet PDFs, downloaded from ST's website, current (21.02.2016) list could 
-  be found in datasheets.txt
-* [pdfminer](https://github.com/euske/pdfminer) tool
 
 ## Running
-If you have the correct XML files, just run `main.py path/to/xml path/to/pdfdir`, 
-where `path/to/Xdir` is the path to a directory containing all xml and pdf files. 
+If you have the correct XML files, just run `main.py path/to/xml`, 
+where `path/to/xml` is the path to a directory containing all xml files. 
 Make sure the XML files have the correct format as there is no error checking.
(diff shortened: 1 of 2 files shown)


SchrodingersGat:
If the VBat pin works better not on the top, then we can ignore this warning. However if this pin should be connected to Vcc then it's probably easier to have on the top.

## Important

If you are reworking these scripts, we **really** need to reduce the length of pin names. It's currently creating ridiculously large parts (A3 size in some cases!)

For pins, they should be named with the Port/Pin number, and then the major function, and potentially one alternate function.

SchrodingersGat:
(It's great that you are improving this script :) )

lorem-ipsum:
I can easily cut the pin names, that should not be an issue. In that case I would cut all the alternate functions as I would argue that the first listed is not always the interesting function.

VBat can be connected to Vcc in case one doesn't want to use a battery. Otherwise it is connected to a battery.

lorem-ipsum:
I've now dropped all the alternate functions. Is there an alternate attribute for pins there they could be stored?

lorem-ipsum:
By shortening the names of all the pins I will also change the size of any schematic. Is that an issue for existing footprints or am I allowed to change all of them?

And if I rework all of them anyways: I would like to split the large ones into multipart components. Every output is 16 pins which makes a nice group size in my opinion. Any thoughts on that?

lorem-ipsum:
I am currently working on the footprint "detection". While doing so I am wondering how to label the CSP footprints. Could you comment on https://github.com/KiCad/Housings_CSP.pretty/issues/8 about that?

SchrodingersGat:
> And if I rework all of them anyways: I would like to split the large ones into multipart components. Every output is 16 pins which makes a nice group size in my opinion. Any thoughts on that?

I would like to see an example of what you are thinking here before commenting.

SchrodingersGat:
> I've now dropped all the alternate functions. Is there an alternate attribute for pins there they could be stored?

Unfortunately this feature does not yet exist. Hopefully in a future version of KiCad this will be possible.

poeschlr:
What is the state of this?

lorem-ipsum:
It is in a state where it is better than the previous script. However it still requires some changes before it can be used to regenerate the full `MCU_ST_STM32` library.

TODOs are:
- [ ] adjust packages and filters to the new layout
- [ ] split the symbols into smaller units.

For splitting the units I was thinking of one unit per IO bank + a central unit with power, boot, clock, etc. pins. This however is a major change which requires more work. I was thinking of using [kipart](https://pypi.python.org/pypi/kipart) for the actual footprint generation.

Ratfink:
Replaced by #212.

PR #147, opened 2017-08-23
Added functionality to add tags as well as checking out specific tags

poeschlr:
This addition adds the following features:
- add a tag (tag given as option.)
- push a tag (option for the tag command)
- checkout a tag (or a branch would work as well)
- cloning using ssh instead of https protocol. (Necessary for push command. Otherwise password entry is necessary for each repo)

Tested with the recent adding of tags to all footprint libs.
I also checked if it can checkout other tags. Works as well.

Tested in linux (fedora 26)

### DIFF
--- a/download_pretty_libs.py
+++ b/download_pretty_libs.py
@@ -14,6 +14,7 @@ import zipfile
 
 # Github address information
 GITHUB_URL = "https://www.github.com/KiCad"
+GITHUB_URL_SSH = "git@github.com:KiCad"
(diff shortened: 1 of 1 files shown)


evanshultz:
@poeschlr 
A few minor things when reviewing the code:
- Line 36: "Checkout" should be "Check out"
- Line 37: Perhaps capitalize "SSH" and "URL"
- Lines 32-37: Should we use the capitalized "GitHub" instead of all lowercase?
- Line 121: Typo of "Tagging" (there are two "g")

I didn't test the code myself.

SchrodingersGat:
Functionally the code is fine, so I'll merge.

Issue #143, opened 2017-08-05
schlib/autogen holds symbol generators but is in git ignore


poeschlr:
Why is the schlib/autogen folder in gitignore?
It currently holds symbol generator scripts.
Are the scripts in the wrong directory or is the gitignore file wrong?

poeschlr:
@jkriege2 or @SchrodingersGat does either of you know something about this?

SchrodingersGat:
I think this is a mistake. The autogen folder is where the scripts should be, but the *output* of these scripts (the generated files) should be ignored.

poeschlr:
The gitignore file has been updated

PR #148, opened 2017-08-24
Connector symbol generator and symbol generator module

poeschlr:
This pull request adds the generator script for the connector symbols added in pull request https://github.com/KiCad/kicad-library/pull/1561

It also adds a module that should help generating symbols. (Sadly i don't have enough time right now to document it. I will do this later.)

As mentioned in https://github.com/KiCad/kicad-library-utils/issues/143 i'm a bit unsure if i put them in the right place. It they are in the right place the gitignore file is clearly wrong.


SchrodingersGat:
@poeschlr great work here. I haven't had time to test it all but I'm happy to merge on the assumption that if any issues crop up in the future they will be fixed then.

PR #152, opened 2017-09-01
rule7_3 fix: fixed module.reference

### LINKED ISSUE #151: rule7_3.py crashes on fix
if rule7_3.py tries to fix it crashes as there is no module.ref property:
[line 249](https://github.com/KiCad/kicad-library-utils/blob/0e7991894b5c943e645fb285774a176310d630a0/pcb/rules/rule7_3.py#L249)
```python
        if self.check():
            if self.refDesError:
                ref = self.module.ref
                if self.checkReference():
```
should be
```python
        if self.check():
            if self.refDesError:
                ref = self.module.reference
                if self.checkReference():
```

s-light:
fixes #151 
the withe space correction was done automatically by my editor (atom).
i can manually change this to not happen - if needed.
pleas let me know

### DIFF
--- a/pcb/rules/rule7_3.py
+++ b/pcb/rules/rule7_3.py
@@ -12,34 +12,34 @@ class Rule(KLCRule):
     """
     def __init__(self, module, args):
         super(Rule, self).__init__(module, args, 'Rule 7.3', "Silkscreen requirements")
-        
+
     def checkReference(self):
         """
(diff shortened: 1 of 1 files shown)


jkriege2:
Hi!

looks good to me ... Thanks,
JAN

PS: In future: Such whitespace-fixes make it hard to spot the actual changes in the code ... so if they are not necessary, I would leave them out ...

PR #156, opened 2017-10-06
Observe correct ordering when saving library

SchrodingersGat:
-  Uses alphabetical ordering of symbol names
- Previously library items were saved in whatever order they were stored


evanshultz:
Looks good to me.

PR #158, opened 2017-10-27
Do not check silk over NPTH or connect pins

evanshultz:
https://github.com/KiCad/Connectors_JST.pretty/pull/52 suffed from false negatives, where a KLC violation was reported when silk was over a NPTH mounting hole. This should be OK, and we agreed to merge the footprints.

This patch fixes that rule so that no checking is done for NPTH- or connect-type holes.

While this patch fixes that particular issue, I'm not very familiar with the check script code so please advise if there are other places this should apply or if there are other concerns.

### DIFF
--- a/pcb/rules/rule7_3.py
+++ b/pcb/rules/rule7_3.py
@@ -119,78 +119,79 @@ class Rule(KLCRule):
                             self.intersections.append({'pad':pad, 'graph':graph})
             else:
                 for pad in module.pads:
-                    padComplex = complex(pad['pos']['x'], pad['pos']['y'])
-                    padOffset = 0 + 0j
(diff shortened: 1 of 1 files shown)


evanshultz:
The diff check triggers off of unchanged code, making it hard to see what actually changed. I only added line 122 which is `if (pad['type'] not in ['np_thru_hole','connect']):`, checking the `else` clause of `checkIntersections()`. Everything after line 122 was not changed (but indented where needed).

evanshultz:
@SchrodingersGat Can you take a look at this PR?

SchrodingersGat:
Seems to do what it says on the tin. Thanks for the reminder :)

PR #164, opened 2017-11-05
Footprint renamer

SchrodingersGat:
This PR is a re-push of the footprint renaming script, which is now more powerful and has guaranteed 57% fewer bugs


poeschlr:
Well if it is guaranteed.

PR #165, opened 2017-11-05
fixed up rule F9.3 (fixmore(), some default fix()-actions, ...)

jkriege2:
This reworks rule F9.3 (http://kicad-pcb.org/libraries/klc/#F9.3) with these features:
- if no model is provided at all: add default model in fix() unless this is a virtual component, which does not always require a 3D model (still all checks are done for virtuals, if models are specified!)
- if errors in 3D model path are there, --fixmore will replace the model path with the default path, which should fix most errors ... exceptions (warning-only), if the model-name starts with the default name, but has an extension (this should capture the e.g. SOT-23_LargePads which should have the same model as SOT-23).

+ for the --fixmore i made a default way to require thise fixes also in other rules (simply set self.needsFixMore=True and implement fixmore() ).

What do yout think?

JAN

### DIFF
--- a/pcb/check_kicad_mod.py
+++ b/pcb/check_kicad_mod.py
@@ -107,7 +107,13 @@ for filename in files:
             printer.yellow("Violating " + rule.name, indentation=2)
             rule.processOutput(printer, args.verbose, args.silent)
 
-        if error:
+        if args.fixmore and rule.needsFixMore:
+            if error:
+                n_violations += 1
+            rule.fixmore()
+            rule.fix()
+            rule.processOutput(printer, args.verbose, args.silent)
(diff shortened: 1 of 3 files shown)


SchrodingersGat:
This looks good but I don't think that you need to issue a warning if a virtual component does not have a 3D model?

I think if a virtual component doesn't have a 3D model the output should be silent...

jkriege2:
I stepped back to warning, so the user is informed that there is not 3D model in that virtual (maybe he wants one ... but I can't think of a use-case right now ...).

SchrodingersGat:
Ok, I'm happy with this, thanks Jan

PR #167, opened 2017-11-07
Fixes for illegal chars

SchrodingersGat:
- Allow '+' char
- Allow ',' char

@poeschlr


poeschlr:
I looked into it and i can't find any reference to these chars being problematic.
Even the microsoft site does not list them:
https://msdn.microsoft.com/en-us/library/windows/desktop/aa365247(v=vs.85).aspx

(But in the wikipedia list they appear. Maybe they where a problem on very old systems.)

I also tested files containing this name on different operating systems and file systems (ext, fat and ntfs, samba share)

So i think this is ok to merge. Thanks.

PR #168, opened 2017-11-10
Added option to save KLC error logs to file

SchrodingersGat:
This PR adds function to save error messages to a log file (JSON)

Required for an idea I have to link users to the appropriate KLC page on the KiCad website when they submit a PR


evanshultz:
I didn't run it, but the logic looks OK to me.

PR #169, opened 2017-11-10
Improve G1.1 using regex

antoniovazquezblanco:
I was playing around to see how things work. This should be a little faster than previous solutions.


poeschlr:
This looks good to me. Sorry it took so long to review it. (Seems to have been missed.)

PR #172, opened 2017-11-14
Fix for --fix

SchrodingersGat:
--fix option previously created noise on ALL files (even those without errors)
Now it only saves files that have violations
Still room for some file noise, but much less


poeschlr:
Thanks

PR #173, opened 2017-11-14
Remove requirement to return 'true' or 'false' in check()

SchrodingersGat:
- Error count incremented when error() is called
- Warning count incremented when warning() is called
- Also added --nowarning option to hide warning output

Currently each rule must return "true" or "false" from the check() function based on whether an error was thrown. 

This is tedious and prone to error. Currently there are a number of rules where errors have been changed to warnings but still return non-zero error codes.

Now, the base Rule class contains an error count and a warning count which are auto-incremented when error() and warning() are called.

### DIFF
--- a/common/rulebase.py
+++ b/common/rulebase.py
@@ -94,18 +94,43 @@ class KLCRuleBase(object):
     def __init__(self, description):
         self.description = description
         self.messageBuffer=[]
+        self.resetErrorCount()
+        self.resetWarningCount()
+
+
+    def resetErrorCount(self):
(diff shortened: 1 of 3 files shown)


SchrodingersGat:
I have also added a `--nowarning` option which allows user to hide warning output if only errors are of interest

evanshultz:
The changes seem OK but I didn't run it. Is it too risky to trust you? :)

SchrodingersGat:
I have run the changes locally but we should be alert for Travis crashes

PR #175, opened 2017-11-18
Sym fp fix

SchrodingersGat:
This PR provides a pretty complicated script to fix footprint associations for symbols.

It allows remapping of:
* Incorrect library names
* Incorrect footprint names
* Footprint provided without a library name

I have tested this extensively and it works well. This will be a necessary script for the transfer to v5 libs as we have changed a LOT of the footprints, both the library structure and the naming scheme!!

Check out the main script file, it has a better description of what it does.

I have started to compile the JSON file with the data we will need to fix the footprint associations. If you run the script with the `--interactive` flag then the JSON file gets updated as you go, DRY principle in action :)


poeschlr:
Thanks for your continuous work on the scripts.

PR #179, opened 2017-11-21
bugfix: F5.3 used the wrong layer-names for silkscreen

jkriege2:
the scripts did not create a correct courtyard on old footpritns without drawing on F.Fab!


SchrodingersGat:
Thanks!

jkriege2:
wow below 60s ... that was quick ;-)

PR #181, opened 2017-11-22
Added script to check 3D model naming

SchrodingersGat:
Simple script that helps determine where 3D models are improperly named.

- Checks folder names
- Checks names of models within folders

Currently using this script to tackle the 3Dpackages naming


evanshultz:
Didn't run it locally, but it looks nice.

PR #182, opened 2017-11-26
fix rule F5.2 for virtual components

jkriege2:
virtual components do not require REFDES and drawing on F.Fab (they do not need to be assembled)


SchrodingersGat:
Good idea, thanks :)

PR #185, opened 2017-11-27
Add footprint checker to S5.1

SchrodingersGat:
- Pass path to footprint files
- S5.1 checks if the specified footprint library or file actually exists

This addition can be used in conjunction with Travis to download the current footprint library and check if the specified footprint exists.

A missing **library** is an error. A missing **footprint** is just a warning (maybe the footprint hasn't been submitted yet)?


evanshultz:
Nice!

PR #186, opened 2017-11-27
Extra checks for lib-table

SchrodingersGat:
- Check duplicate entries
- Check improper paths
- Check mislabeled libs

This PR adds extra checks for lib-table check script


poeschlr:
thanks

PR #187, opened 2017-11-27
F5.2fix

SchrodingersGat:
Recent bug introduced an error in F5.2. An "if" statement ate the check to test if the second RefDes was null.


evanshultz:
Thank you!

PR #192, opened 2017-11-28
allow roundrect pads, which will be available in KiCAD v5

jkriege2:


SchrodingersGat:
I'm happy with this but we should not merge until we mark the v4 libs as archived, as those libs are currently served by these scripts also.

pointhi:
in v5 we get polygone pads as well

jkriege2:
Hi!

I added "custom" to the allowed pad shpaes ... now all available pad shapes are allowed, so should we maybe remove that rule altogether?

JAN

SchrodingersGat:
Yep, let's remove it, especially given it is a "hidden" (undocumented) rule.

jkriege2:
is removed ... can be merged?

SchrodingersGat:
Thanks :)

PR #194, opened 2017-12-05
Add option to generate shielded connector symbols. +bugfix

poeschlr:
Symbols are added in https://github.com/KiCad/kicad-symbols/pull/26


poeschlr:
@jkriege2 could you take a look at this?

PR #197, opened 2017-12-26
Calculate center of symbol outline for origin

gauravjuvekar:
instead of center of all pins. This is done only if the symbol has a single rectangle with foreground fill.

### DIFF
--- a/schlib/rules/S3_1.py
+++ b/schlib/rules/S3_1.py
@@ -26,28 +26,18 @@ class Rule(KLCRule):
             x = (int(rect['startx']) + int(rect['endx'])) // 2
             y = (int(rect['starty']) + int(rect['endy'])) // 2
         else:
-            x_min = y_min = x_max = y_max = None
-
             pins = self.component.pins
 
             # No pins? Ignore check.
             # This can be improved to include graphical items too...
             if len(pins) == 0:
                 return False
-
-            for i, p in enumerate(pins):
-
-                x = int(p['posx'])
-                y = int(p['posy'])
-
-                if i == 0:
-                    x_min = x_max = x
-                    y_min = y_max = y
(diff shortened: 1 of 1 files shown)


jkriege2:
Hi!

could you explain this PR a bit more? What's the rationale behind it?

gauravjuvekar:
KLC S3.1 http://kicad-pcb.org/libraries/klc/#_symbol_guidelines says that symmetrical symbols should be centered around the origin.

I think that this is a little bit ambiguous as to what constitutes the symbol. Is the main rectangular body considered the symbol, or just the ends of all pins? From the example image in the KLC, the origin is the center of the rectangular body of the symbol, not the "enclosing rectangle" of all pin origins.

For example consider this image.

selection_412

IMO, this is a correct symbol - the origin is at the center of the rectangular body. However, the script warns that it violates S3.1
```
Checking symbol 'TPD2EUSB30':
  Violating S3.1
    Origin is be centered on the middle of the symbol
    Symbol not centered on origin
     - Center calculated @ (0, -200)
  Violating S3.6
    Pin name position offset
    Pin offset ouside allowed range
     - Pin offset (0) should not be below 20mils
```
because of the long GND pin at the bottom which shifts down the center of "containing rectangle of pin positions" down from the center of the symbol "body rectangle".

If this reasoning is correct, we should also probably reword the KLC section a bit.

gauravjuvekar:
@poeschlr @jkriege2 @SchrodingersGat @evanshultz 
Bump.

jkriege2:
I think this makes sense ...

@SchrodingersGat @poeschlr @evanshultz What do you think?

SchrodingersGat:
Yep this approach makes sense to me :)

PR #198, opened 2018-01-01
added check for mismatch between the footprint field contents and the filters

jkriege2:
I noticed that for several submissions we weither habe only the footprint field or only filters set, but not both as required for S5.1/5.2. So I came up with this check for these problems ... 
- too many/too few filters
- single-filter but no footprint field ...
- check FPFilter(s) against provided default footprint

Note: This might lead to errors for generic components like LEDs with `LED*` FPFilter only, but these cause several errors anyways!

### DIFF
--- a/schlib/rules/S5_1.py
+++ b/schlib/rules/S5_1.py
@@ -1,6 +1,7 @@
 # -*- coding: utf-8 -*-
 
 from rules.rule import *
+import fnmatch
 
 class Rule(KLCRule):
     """
@@ -73,6 +74,10 @@ class Rule(KLCRule):
                                     self.error("Specified footprint does not exist")
                                     self.errorExtra("Footprint file {l}:{f} was not found".format(l=fp_dir, f=fp_path))
                                     
+                    for filt in filters:
+                        if (not fnmatch.fnmatch(fp_path, filt)) and (not fnmatch.fnmatch(fp_name, filt)):
+                            self.error("Footprint filter '"+filt+"' does not match the footprint '"+fp_name+"' set for this symbol.")
+                            fails=True
                 if len(filters)==0:
                     self.error("Symbol has a footprint defined in the footprint field, but no footprint filter set. Add a footprint filter that matches the default footprint (+ possibly variants).")
                     fails=True
--- a/schlib/rules/S5_1.py
+++ b/schlib/rules/S5_1.py
@@ -26,6 +26,8 @@ class Rule(KLCRule):
                 fp_name = fp_name[1:-1]
 
             fp_desc = "Footprint field '{fp}' ".format(fp=fp_name)
+            
+            filters = self.component.fplist
 
             # Only check if there is text in the name
             if len(fp_name) > 0:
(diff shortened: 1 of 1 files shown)


SchrodingersGat:
I'm not sure if I agree with the logic here, I believe we should have the following options:

1. No footprint, zero or more filters
   * No warnings or errors issued, e.g. generic parts
1. Footprint, no filters
   * Provide a warning "add filter(s) to match possible footprints"
   * Not an "error" - if there is only one possible footprint is does not need to be filtered as it is already assigned. 
1. Footprint, one or more filters
   * Filters should match default footprint (warning if they do not)
1. No footprint, one or more filters
   * No warnings or errors issued (beyond what is already tested)

jkriege2:
I would see it a bit different:

1. No footprint, zero filters: Even generic parts usually have footprint filters, 
2. No footprint, one filter: Probably this is a user-error (didin't assign footprint field, only set filter), so I would like to warn the user about this (at least). Most of the symbols  submitted are not generic, so a warning would help most comitters, I think. In my view all typical symbols should have footprint/FPFilter-info provided.
3. no footprint, several filters: no warning probably a multi-footprint symbol
4. footprint, no filter: After the current KLC this seems to be an error: 

> Footprint filters are used to help match appropriate footprints to a given symbol. This is important even for atomic symbols (symbols with only one matching footprint), as there still may be multiple compatible footprints with different variants or options

5. footprint, 1 filter: alle fine, check if filter matches footprint and warn if not
6. footprint, several filters: footprint has been given, so this is assumes to be an atomic part, i.e. there should only be one filter ... so I would see this as an error (either footprint field assignment is wrong, or too many filters, as what is the point in several filters matching one footprint that could at most have some variants?

About what's a warning and what's an error: I tend to make more things errors, as people will not really see the warning, when using the online-check, because it appears "green" on the main page (usually people then think all is fine and skip checking the output) ... maybe one could get green (no errors), yellow (warnings only) and red (errors) for the Travix-box? That would partly solve that problem?

What do you think on my rationale? I'll wait which conclusion we reach and modify code afterwards accordingly.

@evanshultz @poeschlr Do you have input on this?

Best,
JAN

poeschlr:
@jkriege2 i would expect the same.

---

A side note: I have been informed that atomic is the wrong term for symbols that have the footprint field set. A part only is atomic if there is a 1:1 releation between its symbol and footprint. (The symbol has a footprint set that is only used for this one symbol.)
A better term to use here would be "fully specified symbol". (= symbol named via mpn, fp field set but point to a generic footprint that is intended to be used by multiple symbols.)

jkriege2:
@poeschlr: Which version? @SchrodingersGat 's or mine?

poeschlr:
The version by @jkriege2

jkriege2:
Which ones would you make warnings and which ones errors?

SchrodingersGat:
> Which ones would you make warnings and which ones errors?

Here are my suggestions based on your comments above:

> 1.  No footprint, zero filters: Even generic parts usually have footprint filters,

Error. Most symbols need a footprint. In the opposite cases we can simply allow an exception.

> 2. No footprint, one filter: Probably this is a user-error (didin't assign footprint field, only set filter), so I would like to warn the user about this (at least). Most of the symbols submitted are not generic, so a warning would help most comitters, I think. In my view all typical symbols should have footprint/FPFilter-info provided.

Error again, I think. I agree with your reasoning here.

> 3. no footprint, several filters: no warning probably a multi-footprint symbol

Agreed, no warning

> 4. footprint, no filter: After the current KLC this seems to be an error:

Agreed, this should be an error.

> 5. footprint, 1 filter: alle fine, check if filter matches footprint and warn if not

Yep

> 6. footprint, several filters: footprint has been given, so this is assumes to be an atomic part, i.e. there should only be one filter ... so I would see this as an error (either footprint field assignment is wrong, or too many filters, as what is the point in several filters matching one footprint that could at most have some variants?

This is also an error.

So, in summary, I now think you were correct the first time :p

SchrodingersGat:
I'm happy to merge this, time will tell if if throws too many false errors (I don't think it will).

Thanks Jan :+1:

PR #199, opened 2018-01-08
Fix schlib/check_lib

jkriege2:
This summarizes three smaller bugfixes/improvements. the first one is a major bug, the rest is only minor stuff:
- **schlib/checklib.py did not save the library-file after a --fix-run in some cases (because a global counter was reset to 0)**
- improved error description in S4.4 
- S6.2: improved fixes to KiCAD fields (visibility, copy datasheet link from field to DCM ...)

### DIFF
--- a/schlib/checklib.py
+++ b/schlib/checklib.py
@@ -92,6 +92,8 @@ for libfile in libfiles:
     if len(libfiles) > 1:
         printer.purple('Library: %s' % libfile)
 
+    n_allviolations=0
+    
     for component in lib.components:
 
         #simple match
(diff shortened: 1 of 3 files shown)


poeschlr:
About the datasheet stuff. I think a patch has been merged that does change kicads behavior regarding these fields. (the one in the lib file vs the one in the dcm file)

The discussion was here: https://lists.launchpad.net/kicad-developers/msg31127.html
This is the message that included the first patch: https://lists.launchpad.net/kicad-developers/msg31171.html

jkriege2:
Yes but if I read that correctly, the datasheet-field is filled from the DCM-file, iff the latter is set and the field is empty, when placing the symbol. Correct? If that is so, KLC is fine (field empty, doc file in DCM set). I added the fix here to ensure that datasheet field contents is automatically transfered to the DCM, if people misplace it (i.e. if they filled the datasheet field instead of the DCM-doc file, thet will be fixed by --fix)

jkriege2:
@poeschlr Please merge this, as the n_violations-stuff causes serious problems (script does not run in some cases locally!)

poeschlr:
I seem to have forgotten about that. Thanks jan.

PR #200, opened 2018-01-09
demote warnings for off-center symbols with less than 50mil off-center to simple info-output

jkriege2:
Often symbols are only 25 or 50mil of-center. In these cases the script now only outputs info ... no longer a real warning ... I would even say we can ignore these cases altogether and give no output at all ... What doy ou think?

JAN


SchrodingersGat:
This makes sense, there's no "fix" for this anyway. Thanks.

PR #202, opened 2018-01-20
Add script to list or remove footprints containing kicad4 incompatible pad shapes.

poeschlr:
This should help kicad version 4 users to use at least part of the new library. (Kicad does only say a lib is invalid. It does not say which footprint is at fault.)

### DIFF
--- a/pcb/check_kicad4_incompatible.py
+++ b/pcb/check_kicad4_incompatible.py
@@ -1,0 +1,38 @@
+#!/usr/bin/env python
+
+from __future__ import print_function
+
+import argparse
+
+import sys,os
+import re
+import glob
+
+regex_str = "\(pad\s*[0-9]\s*(?:smd|thru_hole)\s(?:roundrect|custom)"
+regex_pattern = re.compile(regex_str)
+
+parser = argparse.ArgumentParser(description='Checks KiCad footprint libs for compatibility with version 4. Outputs a list of incompatible footprints. (optionally removes them to allow the lib to be used with KiCad version 4)')
+parser.add_argument('libs', nargs='+')
+parser.add_argument('-r', '--remove', help='remove incompatible footprints', action='store_true')
+
+args = parser.parse_args()
+
+invalid = []
+
+for lib in args.libs:
+    if not lib.endswith(os.sep):
+        lib += os.sep
+
+    #print(glob.glob('{:}*.kicad_mod'.format(lib)))
+    for fp in glob.glob('{:}*.kicad_mod'.format(lib)):
+        with open(fp, 'r') as fp_file:
+            for line in fp_file.readlines():
+                if re.search(regex_pattern, line):
+                    invalid.append(fp)
+                    break
+
+print('\n'.join(invalid))
+
+if args.remove:
+    for file in invalid:
+        os.remove(file)
--- a/pcb/check_kicad4_incompatible.py
+++ b/pcb/check_kicad4_incompatible.py
@@ -8,7 +8,7 @@ import sys,os
 import re
 import glob
 
-regex_str = "\(pad\s*[0-9]\s*(?:smd|thru_hole)\s(?:roundrect|custom)"
+regex_str = "\(\s*pad\s*[^\s]+\s*(?:smd|thru_hole)\s*(?:roundrect|custom)"
 regex_pattern = re.compile(regex_str)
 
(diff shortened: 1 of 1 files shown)


jkriege2:
Sounds like a great idea!

I would (though) change your RegExp a bit: from `\(pad\s*[0-9]\s*(?:smd|thru_hole)\s(?:roundrect|custom)` to `\(\s*pad\s*[^\s]+\s*(?:smd|thru_hole)\s*(?:roundrect|custom)`
This way it is more robust to additional whitespaces and does not require exactly one digit for pin-number, but any non-whitespace-sequence if OK (e.g. also A23 or something like that)

jkriege2:
PS: Should we have the same for the schematics, i.e. automatically change version 2.4 to 2.3?

poeschlr:
Updated the regex

poeschlr:
@jkriege2 could you take another look at this?

poeschlr:
@jkriege2, @evanshultz or @Ratfink could one of you take another look at this?

And yes i plan something similar for the symbol lib as well. (I will use this to generate the kicad 4 backport release of this lib.)

Ratfink:
I'll try to take a look tomorrow.

pcb/check_kicad4_incompatible.py
```diff
@@ -0,0 +1,38 @@
...
+args = parser.parse_args()
+
+invalid = []
+
+for lib in args.libs:
+    if not lib.endswith(os.sep):
+        lib += os.sep
+
+    #print(glob.glob('{:}*.kicad_mod'.format(lib)))
+    for fp in glob.glob('{:}*.kicad_mod'.format(lib)):
```
Ratfink: Use `glob.iglob` to avoid having to store all the filenames at once.

Also, you could replace the string formatting and `if` block above with a single call to `os.path.join`, like this:
```python
for fp in glob.iglob(os.path.join(lib, '*.kicad_mod')):
```

pcb/check_kicad4_incompatible.py
```diff
@@ -0,0 +1,38 @@
...
+parser = argparse.ArgumentParser(description='Checks KiCad footprint libs for compatibility with version 4. Outputs a list of incompatible footprints. (optionally removes them to allow the lib to be used with KiCad version 4)')
+parser.add_argument('libs', nargs='+')
+parser.add_argument('-r', '--remove', help='remove incompatible footprints', action='store_true')
+
+args = parser.parse_args()
+
+invalid = []
+
+for lib in args.libs:
+    if not lib.endswith(os.sep):
```
Ratfink: I suggest removing this `if` block, see below.

pcb/check_kicad4_incompatible.py
```diff
@@ -0,0 +1,38 @@
...
+        lib += os.sep
+
+    #print(glob.glob('{:}*.kicad_mod'.format(lib)))
+    for fp in glob.glob('{:}*.kicad_mod'.format(lib)):
+        with open(fp, 'r') as fp_file:
+            for line in fp_file.readlines():
```
Ratfink: To avoid reading the entire file at once and storing it in memory as a list of strings, just do:
```python
for line in fp_file:
```

pcb/check_kicad4_incompatible.py
```diff
@@ -0,0 +1,38 @@
+#!/usr/bin/env python
+
+from __future__ import print_function
+
+import argparse
+
+import sys,os
```
Ratfink: `sys` is unused.

evanshultz:
Functionally I have nothing more to add than what's above. However:
1. There is no docstring or other general file information. (It's fairly straightforward, and the argparse interface provides some self-documentation, but just as a formality...)
1. When printing any v4-incompatible footprints on line 34, perhaps the lib name should be printed too? Since args.libs can be a list of many libraries, it saves the user from searching to find in what lib the offending footprint lives. Sometimes this might be fairly obvious, but perhaps not all the time.

Ratfink:
I agree that it would be nice to add a module docstring, especially if the script were refactored as a module that can be executed rather than merely a script.

`invalid` contains the whole (absolute or relative) paths to the non-v4-compatible footprints, so where they're printed on line 34, the library name already does get included.

Ratfink:
@poeschlr Do you think you'll be able to finish this up soon?

Ratfink:
Well, all right.  It works, and you're right that this won't be run often, so there's no reason to force the issue.  Thanks!

PR #203, opened 2018-01-21
improved error-reporting on S5.1

jkriege2:
gives extra output when FPFilter does not match footprint, which should better explain where the problem is!


poeschlr:
Looks good. Thanks

PR #205, opened 2018-01-26
Add script to rename 3d models

poeschlr:
This does use the json file that would normally be used by the
fix_footprints script for fixing footprint fields of symbols.

### DIFF
--- a/packages3d/rename.py
+++ b/packages3d/rename.py
@@ -1,0 +1,56 @@
+#!/usr/bin/env python
+# -*- coding: utf-8 -*-
+
+from __future__ import print_function
+
+import argparse
+import sys, os
+import re
+import json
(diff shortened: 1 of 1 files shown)


Ratfink:
I agree with you for scripted models, but not all of our 3D models are scripted.  I think this script could be useful for renaming models that were made by hand, but it would need to rename the .FCStd file as well, which it doesn't do in its current form.  Also, STEP files have the 3D model name in them multiple times, and I don't know how important it is to keep this up-to-date with the new filename.  This is really what I was asking about before.

poeschlr:
This script has never been intended to replace maintenance of the scripted 3d models. (It got used to speed up the process of fixing the lib for the v5 release.)

PR #206, opened 2018-01-27
PCB-check-utily now also saves models, if --fixmore correct some warnings

jkriege2:
`--fixmore` can be used to fix some warnings (specifically 3D model paths/names). In those cases the PCB-check untility didn't always save the modified model. This PR fixes that problem.

### DIFF
--- a/pcb/check_kicad_mod.py
+++ b/pcb/check_kicad_mod.py
@@ -128,11 +128,15 @@ for filename in files:
         if args.fixmore and rule.needsFixMore:
             if rule.hasErrors():
                 n_violations += rule.errorCount
+            if rule.hasWarnings:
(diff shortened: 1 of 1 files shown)


poeschlr:
Do i read this change correctly that until now only --fix did that fix and you added the functionality that now also fixmore does the same?

jkriege2:
The functionality is: Some fixes we don't want to do with --fix, as they might break an intentional thing (like 3D model pathes or auto-courtyard). 
The problem was that the fixmore-fixes were doe with the option, but the changed model was never saved, so they were never written to harddisk (for --fix that worked as expected). This PR solves that problem

poeschlr:
Ah ok. Thanks for the clarification. I assume you tested it so i will merge it as is.

PR #207, opened 2018-01-28
added heuristics to recognize "small" symbols (resistor, transistor, diodes, ...) and modify KLC-requirements accordingly

jkriege2:
This PR adds a heuristics to detect "small" components (e.g. resistors, diodes, transistors), currently with these rules:
1. 2-pins are always small components (most ICs have more pins)
2. 3 or 4 pins and no surrounding symbol outline that is filled with "BACKGROUND" (typically used for symbol outlines according to KLC)

When this heuristics thinks it's a small component I modify a few KLC rules:
- S3.3 outputs only a warning if it does not find a surrounding rectangle with 10mil+BACKGROUND-filling, instead of an error (e.g. we don't want errors with resistors that actually might have such a rectangle)
- S4.1 relaxes its conditions (usually: require 100mil grid, error for pins <50mil and warning <100mil) to require 50mil-grid, error for pin-lengths <25mil and a warning for pin-lengths <50mil

This way we get rid of several warnings/errors e.g. in Transistor_*.lib, Dveice.Lib, ... but retain errors for e.g. 3-pin temperature-sensors that are drawn with a rectangle (I tested on MAX31820)

### DIFF
--- a/schlib/rules/S3_3.py
+++ b/schlib/rules/S3_3.py
@@ -25,11 +25,20 @@ class Rule(KLCRule):
         self.n_rectangles = len(self.component.draw['rectangles'])
         if self.n_rectangles != 1: return False
 
-        if (self.component.draw['rectangles'][0]['thickness'] != '10'):
-            self.error("Component line is thickness {0}mil, recommended is {1}mil".format(self.component.draw['rectangles'][0]['thickness'],10))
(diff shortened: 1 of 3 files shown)


poeschlr:
Looks good to me. Thanks. (I trust you tested it. Otherwise new contributions will be our testcases.)

PR #208, opened 2018-01-29
fixes rule-check for S3.2 (label/name fields of pins)

jkriege2:
- fixes rule-check for S3.2 to also correctly check number text size for the range 20..50mil 
- plus: added improved output on errors/warnings, hinting at the allowed range
- gets check more in line with rule http://kicad-pcb.org/libraries/klc/S3.2/


poeschlr:
Looks good. Thanks

PR #211, opened 2018-02-14
Swap name_text_size and num_text_size

fauxpark:
The order of num_text_size and name_text_size in _PIN_KEYS is wrong. If a symbol violates S3.2, checklib.py outputs the opposite warning.

Example:
For a pin with a 50mil number and 40mil name:
```
X VDD 4 0 400 100 D 50 40 1 1 W
```
You get this:
```
Pin number text size should be 50mils (or 20...50mils if required by the symbol geometry)
```


jkriege2:
Thanks,
JAN

PR #215, opened 2018-03-01
Add support for the new offset field for model placement.

poeschlr:
Newer versions of kicad use the metric 'offset' keyword instead of the imperial 'at' to communicate 3d model offset. This commit does ensure that such newer files can be parsed.

While looking at this i noticed that neither scale, rotation nor offset are checked in any rule. (mention for requirements for rotation and offset are also absent from the KLC rule http://kicad-pcb.org/libraries/klc/F9.3/)


evanshultz:
Looks OK to me just looking at the code.

PR #216, opened 2018-03-01
Add additional checks and fix options for 3d model settigns.

poeschlr:
While creating the fix for the new offset property (PR https://github.com/KiCad/kicad-library-utils/pull/215) i noticed that there are no checks for 3d model scale, rotation and offset.
This PR does add a check and fix option for this.

I also removed duplicate fixmore code. (I assume in the past there was no separate fixmore member function but that was handled as part of the fix function.)

### DIFF
--- a/pcb/rules/F9_3.py
+++ b/pcb/rules/F9_3.py
@@ -10,17 +10,54 @@ class Rule(KLCRule):
     Create the methods check and fix to use with the kicad_mod files.
     """
     def __init__(self, module, args):
-        super(Rule, self).__init__(module, args, '3D model reference')
+        super(Rule, self).__init__(module, args, '3D model settings')
 
     def checkModel(self, model):
 
         error = False
 
+        self.model3D_wrongOffset = False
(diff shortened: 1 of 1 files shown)


evanshultz:
Maybe use "3D model" to be more clear:
image

This looks like debug code:
image

Lastly, the wording on line 201 is awkward at best.

poeschlr:
Debug code removed.

I did not change anything in line 201. In all my tests i never got that message (The reason why there was debug code around was that i tried to find out why this is the case.)

I left it in there in case i missed some edge case where the execution can reach this part.

evanshultz:
Yes, I see you didn't change line 201 but using "fix" so many times is awkward. And since I was reading it now, perhaps it could be changed to "Use --fixmore instead of --fix to correct the 3D model settings!".

Also, when I read it again the wording could be further improved to ""3D model [offset|rotation|scale] is not xxxxxxx".

But I'm OK with this so let me know if you just want me to merge.

poeschlr:
Done

PR #219, opened 2018-03-08
Don't accept STEP files

Ratfink:
KLC F9.3 currently says that only .wrl files are allowed for 3D model associations.  This PR updates the checking script to not accept .step files.

### DIFF
--- a/pcb/rules/F9_3.py
+++ b/pcb/rules/F9_3.py
@@ -54,7 +54,7 @@ class Rule(KLCRule):
                 .format(s=model['scale']))
 
         # Allowed model types
-        extensions = ["step", "stp", "wrl"]
+        extensions = ["wrl"]
 
         model = model['file']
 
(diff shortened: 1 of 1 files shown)


evanshultz:
@poeschlr @Shackmeister 
Please check this. The code changes above are correct if we only want to allow WRL files (which we do, or at least did). Is anything changing? This would have been wrong so I'm unsure why the check script would already be written this way. Why would we have already allowed STEP files?

Ratfink:
@evanshultz [KLC 2.0](https://github.com/KiCad/kicad-library/wiki/Kicad-Library-Convention/c6265cea84aee13aa1b32a3a5745ee93d5932c02) allowed STEP files for a short time.  This was removed by [KLC 2.0.1](https://github.com/KiCad/kicad-library/wiki/Kicad-Library-Convention/06b2c124f5f062e222b4e7925cb19154e5df4ba3), but I guess the script hasn't been updated since.

evanshultz:
Ah! Alright then. Thank you. That means this change is good.

Issue #224, opened 2018-03-13
Escaped characters in strings not parsed properly


Ratfink:
https://github.com/KiCad/kicad-footprints/pull/388 led me to the discovery that the S-expression parsing code doesn't correctly handle escaped characters in strings.  Most significantly, escaped quote marks (`\"`) shouldn't end the string being parsed, but currently they do.  This alone can be fixed with a small change to the `term_regex` expression, specifically the `<sq>` group.

It would be nice to properly decode any escape sequences rather than reading them literally.  I know that `\n` can also occur in KiCad's S-expressions, representing a newline.  The documentation for the file formats isn't especially thorough, so there may be others as well.

PR #225, opened 2018-03-13
Properly handle escaped quotes in s-expressions

Ratfink:
This fixes the first part of #224 by not ending the parsing of a quoted string when \\" is seen.


Ratfink:
Fixed!  Now when we read escaped quotes inside strings, they get loaded as literal quotes.  When writing S-expressions, quotes in strings get escaped properly.

poeschlr:
Looks ok, thanks

PR #226, opened 2018-03-14
MPN and Datasheet assingment tool

cyber-murmel:
I wrote a python script that allow the fast addition/change of the MPN field of every component of a schematic. The goal is to enable fast and consistent changes. This enables for example the switch from one part supplier to another without manually changing the schemaic.

The script iterates over all components of a schematic and tries to find 100% matches in the supplied lookup table json file. A lookup table might look like this:
```
{
    "C": [
        {
            "Value"             : "100n",
            "Footprint"         : "Capacitors_SMD:C_0402",
            "Tolerance"         : "5%",
            "MPN"               : "12345",
            "Datasheet"         : "https://www.url.tld/12345.pdf"
        },
        {
            "Value"             : "1u",
            "Footprint"         : "Resistors_SMD:R_0402",
            "Tolerance"         : "10%",
            "MPN"               : "abcde",
            "Datasheet"         : "https://www.url.tld/abcde.pdf"
        }
    ],
    "R": [
        {
            "Value"             : "100",
            "Footprint"         : "Capacitors_SMD:C_0402",
            "Tolerance"         : "5%",
            "MPN"               : "67890",
            "Datasheet"         : "https://www.url.tld/67890.pdf"
        },
        {
            "Value"             : "1k",
            "Footprint"         : "Resistors_SMD:R_0402",
            "Tolerance"         : "10%",
            "MPN"               : "fghij",
            "Datasheet"         : "https://www.url.tld/fghij.pdf"
        }
    ]
}
```

### DIFF
--- a/README.md
+++ b/README.md
@@ -31,7 +31,44 @@ KiCad utilities
 
 **update_footprints.py**: This script updates the footprints fields of sch files using a csv as input.
 
-**assign_mpns.py**: This script updates the MPN and Datasheet fields of sch files using a json as input.
+**assign_mpns.py**: This script updates the MPN and Datasheet fields of sch files using a json lookup table as input.
+The json file may look like this
+```json
+{
+    "C": [
+        {
+            "Value"             : "100n",
+            "Footprint"         : "Capacitors_SMD:C_0402",
+            "Tolerance"         : "5%",
+            "MPN"               : "12345",
+            "Datasheet"         : "https://www.url.tld/12345.pdf"
+        },
+        {
+            "Value"             : "1u",
+            "Footprint"         : "Resistors_SMD:R_0402",
+            "Tolerance"         : "10%",
+            "MPN"               : "abcde",
+            "Datasheet"         : "https://www.url.tld/abcde.pdf"
+        }
+    ],
+    "R": [
+        {
+            "Value"             : "100",
+            "Footprint"         : "Capacitors_SMD:C_0402",
+            "Tolerance"         : "5%",
+            "MPN"               : "67890",
+            "Datasheet"         : "https://www.url.tld/67890.pdf"
+        },
+        {
+            "Value"             : "1k",
+            "Footprint"         : "Resistors_SMD:R_0402",
+            "Tolerance"         : "10%",
+            "MPN"               : "fghij",
+            "Datasheet"         : "https://www.url.tld/fghij.pdf"
+        }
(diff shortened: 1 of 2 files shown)


Ratfink:
Can't this be done already using the "Edit Symbol Fields" dialog in Eeschema?

cyber-murmel:
Yeah. One by one. This tool is about mass MPN assignment by matching values.

Ratfink:
I might be missing what you mean by mass assignment, because as far as I can tell, you *can* do mass assignment by matching values in the dialog I mentioned.  Make sure that "Group By" is checked for exactly the fields you want to match (in your case, it looks like that would be "Value", "Footprint", and "Tolerance"), then set your MPN field for the groups you're interested in.  Is this not essentially the same as what your script does?

cyber-murmel:
I may also be missing something? How do I get to this dialog in KiCad 5?

Ratfink:
Click the toolbar button in Eeschema that looks like a blue and white spreadsheet, inbetween the netlist and BOM buttons:
screenshot from 2018-08-02 17-26-08
Then you get this dialog, which I believe is essentially a more general version of the script in this PR, as well as the others in the sch/ directory of this repository:
screenshot from 2018-08-02 17-27-19

poeschlr:
In kicad 5 tools -> edit fields allows to group by value. you can then change other fields of such a group all at once.
I just made an experiment in my kicad 5 under ubuntu. Activate grouping by value (marked pink in the screenshot. I also grouped by footprint in my example as i want to show that this is even more flexible. This way i can give the 0805 version a different HPN then the 0603 even if the value field is the same.) The entry at the "head" of the group can be used to update every symbol in that group.

screenshot from 2018-08-02 23-31-07


After entering the desired value in the group level the result looks like this:
screenshot from 2018-08-02 23-31-18

cyber-murmel:
Ah, nice. Didn't know that. :+1: 

Is there also a way to run this as a pre commit hook or is it GUI only?

Ratfink:
I don't think it can be accessed from outside the GUI.  Eeschema, unlike Pcbnew, is quite lacking in the scripting department.  I suppose you may still want this script if running this as a Git hook is important to your workflow.

I'd like to defer the question of whether this should be merged or not to #250, as I (personally) don't understand why the librarians should have to maintain scripts that have nothing to do with the libraries.

PR #227, opened 2018-03-14
Fix description of rule S3.1

Ratfink:
Somehow the word "be" crept in where it doesn't belong.  This fixes it.

@evanshultz would you mind taking a look, like I did for your similar PR a couple days ago? 🙂

### DIFF
--- a/schlib/rules/S3_1.py
+++ b/schlib/rules/S3_1.py
@@ -8,7 +8,7 @@ class Rule(KLCRule):
     Create the methods check and fix to use with the kicad lib files.
     """
     def __init__(self, component):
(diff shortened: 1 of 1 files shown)


evanshultz:
Sure. Would it be even better to word it `Origin is at the center of the symbol`?

Ratfink:
This just makes it match the name that KLC currently has for that rule.  http://kicad-pcb.org/libraries/klc/S3.1/

Personally, I'd prefer those texts to be prescriptive rather than descriptive.  So something more like "Symbol should be centered on the origin".  But I think that discussion should be about changing KLC, then we can update the library-utils to match.

evanshultz:
Eh. Alright then. A little wordsmithery everywhere in KLC would be nice when time allows.

Thanks!

PR #228, opened 2018-03-14
Minor reorganization of README.md

Ratfink:
This PR fixes some minor grammatical errors in the README's descriptions of the various scripts.  A few outdated entries are removed.  It also rearranges the listings to put the most commonly used script for each directory first.

Whoops, I accidentally put this branch directly in the main repository.  I'll delete the branch once this PR is closed.

### DIFF
--- a/README.md
+++ b/README.md
@@ -6,11 +6,11 @@ KiCad utilities
 
 **checklib.py**: Script for checking [KLC][] compliance of schematic symbol libraries.
 
-**schlib.py**: A Python module for parsing KiCad's Schematic Libraries Files Format.
+**schlib.py**: A Python module for parsing KiCad's schematic library file format.
 
 **test_schlib.sh**: A shell script used to validate the generation of files of the schlib module.
 
-**fix-pins.py**: A script created to help adapt existing library files to the [KiCad Library Convention][KLC] by testing some cases of x/y "wrong" pin positions and trying to fix them. The description of the cases are explained in the head of the script file.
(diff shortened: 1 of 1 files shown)


evanshultz:
- Does `Schematic Libraries Files Format` need to be in caps? See other examples too.
- For fix-pins.py, should the `head` in `head of the script file.` be `header`?
- For update_footprints.py, should `csv` be capitalized? Actually, file extensions, if they're too be capitalized, can happen in a few other places too.
- kicad_mod.py: "working with" isn't descriptive enough IMO.
- The first sentence for check_3d_coverage.py reads awkwardly to me. It seems far more in-depth than the other file descriptions and the scripts are complex enough to keep the descriptions very generalized. Maybe something like `Script for checking if a KiCad footprint file has a 3D model.`? Keep the second sentence.

evanshultz:
@poeschlr 
Personally, I would prefer if you merge this so you've gotten to check it out and approve it.

Ratfink:
* I guess "Schematic Libraries Files Format" doesn't need to be capitalized.  I believe it's referring to the header in the [file format documentation](https://bazaar.launchpad.net/~stambaughw/kicad/doc-read-only/download/head:/1115%4016bec504-3128-0410-b3e8-8e38c2123bca:trunk%252Fkicad-doc%252Fdoc%252Fhelp%252Ffile_formats%252Ffile_formats.pdf/file_formats.pdf), but I'll change it to lower case because that would look better.
* Sure, "header" does seem better.
* I'd suggest that if we want file extensions to stand out, we should write them like `.csv`.
* Okay, I'll make that more descriptive.
* I'll simplify this sentence as suggested.

poeschlr:
Looks good to me. For the auto gen stuff we might want to either make the description more generic or add information about the other generators that exist.

evanshultz:
Props for using an Oxford comma! Love it!

PR #232, opened 2018-03-31
Add initial support for custom pad shapes

Ratfink:
This PR adds initial support for custom pad shapes in the KicadMod class.  This does most of the work described in #231.  The KicadMod class can now load all custom pad primitives.  It can also check the bounding box of all custom pad primitives except for arcs, which will be more complicated to implement than the others.  Since there are no footprints in kicad-footprints yet that use arc primitives (or in fact, any primitives except for polygons) in custom pads, I don't think this is a problem yet. 🙂

There is no support for exporting custom pad information yet either.

### DIFF
--- a/pcb/kicad_mod.py
+++ b/pcb/kicad_mod.py
@@ -418,6 +418,72 @@ class KicadMod(object):
             a = self._getArray(pad, 'thermal_gap')
             if a: pad_dict['thermal_gap'] = a[0][1]
 
+            # Custom pad shape settings
+            if pad_dict['shape'] == 'custom':
+                # Get options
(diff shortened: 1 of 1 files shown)


poeschlr:
I can not see anything obviously wrong. 
I assume you tested it locally? (I will merge as soon as you confirmed that.)

Ratfink:
Yes, I tested it locally.  It correctly processes https://github.com/KiCad/kicad-footprints/pull/461, and a test footprint I kept adding random primitives to while writing this.

poeschlr:
Oh yes. sorry forgot about that.

PR #235, opened 2018-04-19
S3.6: Allow offset to be 0

Ratfink:
An offset of 0 in the library file is special: it means the pin name is next to the pin rather than inside the symbol.  In some cases, that's the only way to draw the pin name, so there's nothing wrong if we're in that case.  This PR makes the script silent when the offset is set to 0.

### DIFF
--- a/schlib/rules/S3_6.py
+++ b/schlib/rules/S3_6.py
@@ -13,7 +13,12 @@ class Rule(KLCRule):
 
         offset = int(self.component.definition['text_offset'])
 
-        if offset > 50:
+        if offset == 0:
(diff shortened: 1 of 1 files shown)


evanshultz:
Thanks for adding a comment! Self-documentation in the scripts is sorely lacking!

Looks good to me. Just two minor questions:
1. Instead of "next to" in the comment for 0 pin name offset, would it be better to always use "inside" and "outside" to match the UI's terms?
1. And on line 18 change "As the offset" to "As offset = 0" or "As this offset" to be a little more specific?

PR #237, opened 2018-04-22
Issue a warning for missing footprint field

### LINKED ISSUE #234: S5.1 error for single footprint filter and no default footprint
I see many symbols that get hit with an error from Travis of `Symbol has a single fooprint filter string '<single footprint filter>' (i.e. it seems to be intended for a single default footprint only), but the footprint field is empty. Fill footprint field with the correct footprint in the form LIBRARY:FOOTPRINT.`

This is from line 91 of https://github.com/KiCad/kicad-library-utils/blob/master/schlib/rules/S5_1.py, which only occurs if there is no default footprint and just a single footprint filter entry. This is correct in many libraries (such as generic connector symbols) and to me an error seems too strong. Even the wording, by using "seems", isn't forceful. It is a worthwhile situation to catch and report, and a warning would give us that while not complaining in cases where this rule definitely does not apply.

Should this just be a warning and not an error?

Lastly, as you can see above, line 93 has a typo of "footprint" by missing the first "t".

Ratfink:
As pointed out in #234, there are plenty of cases where a symbol having one footprint filter and no default footprint is valid, and therefore this shouldn't be treated as an error by the script.

This commit changes that case to a warning, splits up the message into a warning and a warningExtra, and shortens the message overall by not explicitly saying how to fix the problem if it in fact is a problem.

Fixes #234.

### DIFF
--- a/schlib/rules/S5_1.py
+++ b/schlib/rules/S5_1.py
@@ -88,10 +88,13 @@ class Rule(KLCRule):
                 if len(filters)>1:
                     self.error("Symbol has a footprint defined in the footprint field, but several ({fpcnt}) footprint filters set. If the symbol is for a single default footprint, remove the surplus filters. If the symbol is meant for multiple different footprints, empty the footprint field.".format(fpcnt=len(filters)))
                     fails=True
-            else:
-                if len(filters)==1:
-                    self.error("Symbol has a single fooprint filter string '"+filters[0]+"' (i.e. it seems to be intended for a single default footprint only), but the footprint field is empty. Fill footprint field with the correct footprint in the form LIBRARY:FOOTPRINT.")
-                    fails=True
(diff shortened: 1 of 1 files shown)


Ratfink:
1. 79, as [PEP 8](https://www.python.org/dev/peps/pep-0008/) says.
2. Those are better names, and I'd unpack the tuple returned by str.split on line 47 to make the file three lines shorter.
3. Those lines could be combined, and the two calls to os.path.join could be combined into one.

But I only intended to touch the part of the file that needed to be changed to fix #234.  I could clean up this file now if you'd like though.

evanshultz:
If you checked this locally, then I'll just merge.

Ratfink:
Yes, I've tested it locally, and it works as intended.

PR #238, opened 2018-04-29
Fix dimensions of TFBGA216

Ratfink:
It's 13x13mm, not 16x16mm.  Fixed in the symbols repository in https://github.com/KiCad/kicad-symbols/pull/543, but I forgot to make a PR for the script until just now.


evanshultz:
As shown, for example, at http://www.st.com/resource/en/datasheet/stm32f427vg.pdf.

PR #243, opened 2018-05-24
Move rule descriptions to rule class docstrings

### LINKED ISSUE #220: KLCRule subclasses have unhelpful docstrings
Python classes and functions can have docstrings to describe their functionality.  We currently have docstrings for most of the KLCRule subclasses, but these all say simply:

```python
"""
Create the methods check and fix to use with the kicad lib files.
"""
```

This doesn't describe the rule at all!  Really, that should be in a parent class' docstring, to describe how the parent is used.

Meanwhile, we have a description that we pass to the parent's constructor.  That's documentation, but it's hidden in a place that's unexpected to Python developers.

I propose that we remove the `description` parameter to KLCRule and KLCRuleBase's constructors, and instead read the description as the first line of the class' docstring.  This can be done as a property of KLCRuleBase, so the code to handle this can exist in exactly one place.  Then our rule descriptions can live in an obvious place, and still get printed exactly as they do now, with minimal new code.

Are there any other thoughts on this?  I'll be happy to implement this if other maintainers think it's a good idea.

Ratfink:
As I wrote in #220, the rule class docstrings seem like a more sensible place for rule descriptions than burying them in the `__init__` method.  I offered to make this change a couple of months ago, and have seen no objections, so here it is.

While moving the descriptions, I changed them all to match the one-line rule names in KLC 3.0.11, the current revision.  All the docstrings I touched now come at least reasonably close to following [PEP 257](https://www.python.org/dev/peps/pep-0257/).  Hooray for good code style!

In moving the descriptions out of the constructor, many rule classes no longer need to define their own constructors at all, so overall this reduces how much code we have to maintain.

Fixes #220.

### DIFF
--- a/common/rulebase.py
+++ b/common/rulebase.py
@@ -91,8 +91,8 @@ class KLCRuleBase(object):
         path = "".join(path.split(".")[:-1])
         return path.replace('_', '.')
 
-    def __init__(self, description):
-        self.description = description
+    def __init__(self):
+        self.description = self.__doc__.strip().splitlines()[0].strip()
         self.messageBuffer=[]
         self.resetErrorCount()
         self.resetWarningCount()
--- a/pcb/rules/F5_1.py
+++ b/pcb/rules/F5_1.py
@@ -7,11 +7,7 @@ from rules.klc_constants import *
 import cmath
 
 class Rule(KLCRule):
-    """
-    Create the methods check and fix to use with the kicad_mod files.
-    """
-    def __init__(self, module, args):
-        super(Rule, self).__init__(module, args, "Silkscreen layer requirements")
+    """Silkscreen layer requirements"""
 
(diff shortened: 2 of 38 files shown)


poeschlr:
Good idea. However we have a merge conflict now as i merged other stuff. Could you look into that?

Ratfink:
The conflict has been resolved; can this be merged now?

evanshultz:
@poeschlr 
Do you still want to revisit this? If so, we'll need to merge in master in order to merge this PR.

poeschlr:
It is a good idea in general but i suspect @Ratfink will no longer work on this. Maybe somebody with a lot of time on their hand will continue this effort.

Which is why i marked it as abandoned as users can then search for this label for things they can help with.

PR #247, opened 2018-06-29
Detect footprint-only name fields in F9.3 check

### LINKED ISSUE #246: F9.3: Carve out specific exceptions?
https://github.com/KiCad/kicad-footprints/pull/691

For `_ThermalVias` footprints we use the non-via 3D model since the package drawing is the same (and the package doesn't have vias). This often creates a warning that we know is OK.

Should the check script have an exception for this specific case to reduce noise?

Ratfink:
As Evan pointed out in #246, it would be great if the script that checks 3D model filenames would give better help with regards to the `_ThermalVias` suffix.  I pointed out a few more fields that should have similar treatment, and that it would be better to not only accept if they're missing from the model name, but warn if they're present.  This patch implements the discussed change, along with a bit of other assorted cleanup to the F9.3 check.  It also makes the --fixmore method remove these special fields from the model name, if any are found.

Fixes #246.

### DIFF
--- a/pcb/rules/F9_3.py
+++ b/pcb/rules/F9_3.py
@@ -103,7 +103,7 @@ class Rule(KLCRule):
 
         if not model_file == fp_name:
             # Exception for footprints that have additions e.g. "_ThermalPad"
-            if fp_name.startswith(model_file) or model_file in fp_name or fp_name in model_file:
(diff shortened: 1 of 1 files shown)


poeschlr:
This looks ok to me. Thanks

PR #249, opened 2018-07-23
Warn about non-nominal silk and fab line widths

### LINKED ISSUE #217: F.Fab and *.SilkS line widths - add warnings?
At https://github.com/KiCad/kicad-footprints/pull/280 I realized that there were 0.15mm line widths everywhere but nothing in the Travis output. As it turns out, this is correct.

F.Fab lines are checked if they're within 0.1mm to 0.15mm (inclusive). Silk lines are checked to be either 0.1mm, 0.12mm, or 0.15mm. If the constraints above are met, there is no output from the check script.

KLC (http://kicad-pcb.org/libraries/klc/#_footprint_guidelines) says this is the correct range, but also gives a nominal/recommended value for each line width. There is additional information there detailing why a non-nominal/non-recommended width could be used for silk lines, but nothing for fab lines.

So, should the check scripts generate a warning if a non-nominal/non-recommended line width is used? Not an error, certainly, but just a warning? If so, I can make a PR for that.

BTW, this doesn't apply to the courtyard because there is only one allowed width.

Ratfink:
As pointed out in #217, it's appropriate to give warnings for non-nominal line widths on the silkscreen and fabrication layers.  This PR changes the checking scripts for rules F5.1 and F5.2 to give these warnings.

For an example of the new output, try running this PR's branch on RF_Module.pretty, as some components in that library have quite a few of both new warnings.

Fixes #217.

### DIFF
--- a/pcb/rules/F5_1.py
+++ b/pcb/rules/F5_1.py
@@ -63,10 +63,13 @@ class Rule(KLCRule):
     def checkSilkscreenWidth(self):
         # check the width
         self.bad_width = []
+        self.non_nominal_width = []
 
         for graph in (self.f_silk + self.b_silk):
             if graph['width'] not in KLC_SILK_WIDTH_ALLOWED:
                 self.bad_width.append(graph)
+            elif graph['width'] != KLC_SILK_WIDTH:
+                self.non_nominal_width.append(graph)
 
     """
     Check if any of the silkscreen intersects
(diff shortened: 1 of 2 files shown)


evanshultz:
@Ratfink 
Thanks!

@poeschlr 
This looks OK to me. Would you take a look and merge or give a thumbs up and I will?

poeschlr:
The idea behind the pull request sounds good. I did however not find time to check if the contribution fulfills its promises.

poeschlr:
I found a bit of time and checked this. Does not look completely wrong so i think we can use it.

PR #252, opened 2018-08-16
Using rounded rectangle pads for marking pin 1 should be allowed.

poeschlr:
This adds rounded rectangle pads to the list of allowed pad shapes for pin 1.
I also noticed that currently only parts with more than 2 pins are checked.


evanshultz:
Looks OK, so just going for it...

PR #255, opened 2018-08-23
common/sexpr: remove dead code

lorem-ipsum:
The class is overwritten a few lines later. Also `__init__` can not work as it calls `kwargs`, but the argument is called `kwarg` (without s).


Ratfink:
Oh, good catch, that sure can be removed.  Thanks!

PR #258, opened 2018-08-31
Add option to compare libs: Test symbol for design breaking changes

poeschlr:
This adds a check if symbols have design breaking changes.
Design breaking changes are moved, removed or renumbered pins.
It also differentiates between symbols where only NC pins have been changed or where real pins have been changed.


schlib/comparelibs.py
```diff
@@ -35,6 +35,7 @@ def ExitError( msg ):
 parser.add_argument("-v", "--verbose", help="Enable extra verbose output", action="store_true")
 parser.add_argument("--check", help="Perform KLC check on updated/added components", action='store_true')
 parser.add_argument("--nocolor", help="Does not use colors to show the output", action='store_true')
+parser.add_argument("--design_breaking_changes", help="Checks if there have been changes made that would break existing designs using a particular symbol.", action='store_true')
```
Ratfink: Long options usually use hyphens to separate words, not underscores.  I'd rather see this option called `--design-breaking-changes`.  An option with this name would be stored as `args.design_breaking_changes` automatically, so you wouldn't have to change any other code to do this.

poeschlr:
This needed a manual merge as the shownochanges option was added between the creation of this pull request and now.

I included everything suggested by @Ratfink could somebody take a look at this as i doubt he will have time?

schlib/comparelibs.py
```diff
@@ -144,23 +146,57 @@ def KLCCheck(lib, component):
...
+                        else:
+                            pins_missing += 1
+                        continue
+
+                    if pin_old['posx'] != pin_new['posx']:
+                        if pin_old['electrical_type'] == 'N' and pin_new['electrical_type'] == 'N':
+                            nc_pins_moved +=1
+                        else:
+                            pins_moved += 1
+                    elif pin_old['posy'] != pin_new['posy']:
```
evanshultz: Since you're not checking if pins moved X or Y, you could combine the "posx" and "posy" clauses. Or does it matter?

schlib/comparelibs.py
```diff
@@ -144,23 +146,57 @@ def KLCCheck(lib, component):
...
+                            nc_pins_moved +=1
+                        else:
+                            pins_moved += 1
+
+                if pins_moved > 0 or pins_missing > 0:
+                    design_breaking_changes += 1
+                    printer.light_purple("Pins have been moved, renumbered or removed in symbol '{lib}:{name}'".format(lib=lib_name, name=cmp))
+                elif nc_pins_moved > 0 or nc_pins_missing > 0:
+                    design_breaking_changes += 1
+                    printer.purple("Normal pins ok but NC pins have been moved, renumbered or removed in symbol '{lib}:{name}'".format(lib=lib_name, name=cmp))
```
evanshultz: Could remove the "Normal pins ok but " part to keep this succinct.

poeschlr: Maybe the wording can be improved but i really want to include the fact that normal pins have not been touched. This is to make it absolutely clear that this symbol should not break 90% of the designs that use it (or maybe it is even closer to 100% to be honest.)

schlib/comparelibs.py
poeschlr: Done

schlib/comparelibs.py
poeschlr: An option would be to make it clear in the other output that these are the important pins that have been moved. Not sure how to word this.

evanshultz:
Everything looks fine and I didn't test this thoroughly, but since it's mostly for internal use and you tested it (right?) I'll merge as is if you wish.

However I added some comments inline.

poeschlr:
well this script was used to create the changelog files for both 5.0.1 and 5.0.2

So yes this has been tested already.

evanshultz:
I meant testing the new changes in this PR. I'll just merge this and then if then if somebody comes up with better wording and takes the time to submit a PR we can review it then.

poeschlr:
Yes i tested that as well. (I did run it on 5.0.1 vs 5.0.2 and it produced the same output as the changelog added to version 5.0.2)

PR #260, opened 2018-09-19
Extract footprint-center from F.Fab outline

cp-aquila:
Some footprints have weird pad shapes, making the F6.2 test fail.
To fix this, we can also calculate the center of the F.Fab layer. Of those
two possible center-locations, the one closer to (0,0) is
selected.
If Pads-center and F.Fab-center differ a warning is displayed.

--------------------

This fixes the travis error for this PR https://github.com/KiCad/kicad-footprints/pull/951

### DIFF
--- a/pcb/rules/F6_2.py
+++ b/pcb/rules/F6_2.py
@@ -46,10 +46,10 @@ class Rule(KLCRule):
         THRESHOLD = 0.001
         if abs(x) > THRESHOLD or abs(y) > THRESHOLD:
             self.error("Footprint anchor does not match calculated center of Pads or F.Fab")
-            self.errorExtra("calculated centers for Pads [{xp},{yp}mm]".format(
+            self.errorExtra("calculated center for Pads [{xp},{yp}mm]".format(
                 xp = round(center_pads['x'], 5),
                 yp = round(center_pads['y'], 5)))
-            self.errorExtra("calculated centers for F.Fab [{xf},{yf}mm]".format(
+            self.errorExtra("calculated center for F.Fab [{xf},{yf}mm]".format(
                 xf = round(center_fab['x'], 5),
                 yf = round(center_fab['y'], 5)))
 
--- a/pcb/rules/F6_2.py
+++ b/pcb/rules/F6_2.py
@@ -3,6 +3,7 @@
 from __future__ import division
 
 from rules.rule import *
+from math import sqrt
 
 class Rule(KLCRule):
     """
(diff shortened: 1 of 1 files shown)


evanshultz:
This should apply only to SMT footprints, since THT footprints have the origin at pin 1.

I didn't look at the code, but I'm not sure what you described above is the best approach. Consider this footprint:
image

The center of the pads and fab lines need to be considered together. Either one separately wouldn't work. This might be a nice incremental improvement and I'm not against merging it, just saying that it's not a comprehensive fix to consider only pads or only fab lines.

cp-aquila:
Hello,

It does not apply to non SMD-Footprints, see Line24. I have not touched that :)

But you are right, this is not a comprehensive fix. There are still valid footprints which will fail. But less will fail with that changeset.

Example: the mentioned HDMI-Footprint the check will not fail, but display a warning.
```
$ ../kicad-library-utils/pcb/check_kicad_mod.py Connector_HDMI.pretty/HDMI_A_Contact_Technology_HDMI-19APL2_Horizontal.kicad_mod -vv
Checking footprint 'HDMI_A_Contact_Technology_HDMI-19APL2_Horizontal':
  Violating F6.2
    For surface-mount devices, footprint anchor is placed in the middle of the footprint (IPC-7351).
    Footprint centers of pads and F.Fab do not match
     - Footprint center calculated as Pad: (0.0,0.0)mm F.Fab: (0.0, 6.55)mm
```

cp-aquila:
This has been idle quite some time.
Does anybody want to review/merge it?
I still consider this a small but usefull addition.

pcb/rules/F6_2.py
```diff
@@ -24,19 +25,31 @@ def check(self):
...
         err = False
 
         THRESHOLD = 0.001
-        x = center['x']
-        y = center['y']
+        # select the xy coordinates that are closest to the center
+        diff_pads = sqrt(center_pads['x']**2 + center_pads['y']**2)
+        diff_fab  = sqrt(center_fab['x']**2 +  center_fab['y']**2)
+        x = (center_pads['x'], center_fab['x'])[diff_pads > diff_fab]
+        y = (center_pads['y'], center_fab['y'])[diff_pads > diff_fab]
```
poeschlr: What is the rationale behind this?

cp-aquila: Line 43 & 44 do calculate the euclidic distance of the pads & fab centers to the origin.
Then the one closes to the origin is selected. (45-46)

That the acutal idea of the patch. Look at the center of the F.Fab lines the one calculated from the Pads.
* If they are not the same, display a warning (line 31-37)
* If they are not the same, select the one that is closer to the footprint origin (line 43-36)
* Then check if the one selected above is to far from the origin (lines 48-52)

Ratfink: Just a stylistic point: Python has a conditional operator `true_value if condition else false_value` that would be better to use than subscripting a tuple on lines 45-46.  But since there are two lines with the same condition, using a normal `if` block could be considered as the more [DRY](https://en.wikipedia.org/wiki/Don%27t_repeat_yourself) option.

cp-aquila: tbh. I don't know why I have written it this way. After the question from poeschlr I had to look twice myself. This is a good indication that the code-quality is shitty.
A simple if-statement is way more readable. Will fix.

pcb/rules/F6_2.py
```diff
@@ -15,28 +16,42 @@ def check(self):
...
         if module.attribute != 'smd':
             # Ignore non-smd parts
             return False
 
-        center = module.padMiddlePosition()
+        center_pads = module.padMiddlePosition()
+        center_fab  = module.geometricBoundingBox("F.Fab").center
+
+        if center_pads['x'] != center_fab['x'] or center_pads['y'] != center_fab['y']:
+            self.warning("Footprint centers of pads and F.Fab do not match")
```
poeschlr: I would not call this a warning. A lot of footprints would result in this and it would just confuse users.

cp-aquila: How about moving the warning output into the code-path that displays the error (if there is any). Then the user has the additional info on which two origins were checked.
Otherwise it will be a mystery for the user why exactly the check failed.

cp-aquila:
A good testcase is `LED-L1T2_LUMILEDS` in `LED_SMD` (which should get fixed).

pcb/rules/F6_2.py
poeschlr: Well if the check fails then i would assume your script knows why and can generate a clear output (you could for example word it such that it is clear which options you checked). This warning on its own however might be read as "i need to ensure the centers are equal" which is not the case.

cp-aquila: I spend some time on better wording, but could not come up with something good. 

Actually I don't know why I introduced it in the first place. There is no real use-case for it.
Now one line is shown for each calculated center. This has all the information the user needs. Its also much clearer than the previous version that had all coordinates in one line.

New look:
grafik

pcb/rules/F6_2.py
```diff
@@ -49,9 +41,13 @@ def check(self):
...
-            self.errorExtra("Footprint center calculated as ({x},{y})mm".format(
-                x = round(x, 5),
-                y = round(y, 5)))
+            self.errorExtra("Footprint center calculated as Pad: ({xp},{yp})mm F.Fab: ({xf}, {yf})mm".format(
+                xp = round(center_pads['x'], 5),
+                yp = round(center_pads['y'], 5),
+                xf = round(center_fab['x'], 5),
+                yf = round(center_fab['y'], 5)))
+            if center_pads['x'] != center_fab['x'] or center_pads['y'] != center_fab['y']:
+                self.errorExtra("Footprint centers of pads and F.Fab do not match")
```
poeschlr: This message is still here. It still suggests that the user needs for some reason ensure pads and fab need to be on the same center.

poeschlr:
Thanks looks good now (ignore my last comment made as i seem to have misclicked and looked on an outdated diff)

PR #261, opened 2018-10-04
correct spelling in interface to KiCadSymbolGenerator

dlharmon:
I'm looking to add a few scripts using this and don't want to propagate it further.


evanshultz:
@poeschlr @Ratfink 
Can you take a look? I think you two wrote these.

Why is there a note that the typo must be kept? Is there an external data file (YAML?) that also has the typo? Can they both be fixed together?

Ratfink:
The note about the typo in the STM32 symbol generator was because I didn't want to change more than necessary in the PR that updated that script.  If this correctly changes every occurrence of the typos, both in the package they're defined in and everything that uses it, it should be good.  I haven't looked closely to see if this is in fact the case though.

dlharmon:
I've found another I'd like to fix (Aligment) and pushed it to this branch. 

There's also a place where Horizontal and Vertical appear to be swapped which I need to investigate further.

I've verified all are changed in this repository. Any others I should be aware of?

SchrodingersGat:
Thanks @dlharmon for these fixes

PR #265, opened 2018-10-23
Enable run mode bit for check_lib_table.py and document it in README

seppestas:
Enable the run mode bit for check_lib_table.py. This makes the script runnable and consistent with the runnable scripts in the pcb, sch, and schlib directories.

I also documented check_lib_table.py (and download_pretty_libs.py, for sake of completeness) in the README. I did not enable the run bit for download_pretty_libs.py since I think it's quite deprecated (and IMHO a bad idea to begin with).


SchrodingersGat:
Thanks @seppestas 

```
I did not enable the run bit for download_pretty_libs.py since I think it's quite deprecated (and IMHO a bad idea to begin with).
```

It was the best of a bunch of bad ideas before all the repos were merged together ;)

PR #266, opened 2018-10-30
Use option flag to decide if symbol is a power symbol

seppestas:
This fixes the PWR_FLAG symbol violating [KLC rule S6.2] "Ref(erence) field must be VISIBLE" because it does not use the #PWR reference.

See section 2.3 - Description of a component in the [KiCad file formats specification](http://bazaar.launchpad.net/~stambaughw/kicad/doc-read-only/download/head:/1115%4016bec504-3128-0410-b3e8-8e38c2123bca:trunk%252Fkicad-doc%252Fdoc%252Fhelp%252Ffile_formats%252Ffile_formats.pdf/file_formats.pdf).

[KLC rule S6.2]: http://kicad-pcb.org/libraries/klc/S6.2/


poeschlr:
We might even go one step further. If the reference starts with "#" then it can be invisible. (the option could be used as a second indicator)

PR #267, opened 2018-11-08
Error checking in sch.Component.addField

airwoodix:
Hello,

these commits fix the error checking TODOs on `sch.Component.addField`:
- `name` and `ref` are made compulsory keyword-only arguments
- invalid additional fields are silently ignored

The function interface changed from a single dictionary to keyword based. This has at least two advantages: self-documentation is better and required arguments are explicit.

The `ensure_quoted` function is there because it is more natural to write `comp.addField(name="foo", ref="bar")` than `comp.addField(name='"foo"', ref='"bar"')`.

The call to `addField` in the `add_part_number.py` script is modified accordingly.


poeschlr:
Sorry for the delay, this looks good to me. (To be honest the sch class is not something we take great care of anyway as we do not really need it for our task of maintaining the library.)

PR #268, opened 2018-11-17
S4.3: Some errors were detected yet a 0 code was returned

diegoherranz:
Issue found on https://github.com/KiCad/kicad-symbols/issues/1114

errorExtra() doesn't mean non-zero return code unless preceded by an error().
So, a few errorExtra() calls now using error() instead.

I've also added a second commit with some code style fixes. Let me know if you disagree with any of that and I can change or remove it.
Edit: I've removed that second commit and moved the style fixes to a different PR (https://github.com/KiCad/kicad-library-utils/pull/269) since they are separate things. It also includes fixes on many other files now.

Thanks!

### DIFF
--- a/schlib/rules/S4_3.py
+++ b/schlib/rules/S4_3.py
@@ -155,7 +155,7 @@ class Rule(KLCRule):
                             self.component.padInSpecialPowerStack.add(pin['num'])
                             # check if all passive pins are invisible
                             if pin['electrical_type'] == 'P' and (not pin['pin_type'].startswith('N')):
-                                self.errorExtra("{pin} : {etype} should be invisible (power-pin stack)".format(
+                                self.error("{pin} : {etype} should be invisible (power-pin stack)".format(
                                     pin=self.pinStr(pin),
                                     etype=pinElectricalTypeToStr(pin['electrical_type'])))
                                 err = True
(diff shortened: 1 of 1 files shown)


diegoherranz:
Gentle reminder.
This can improve the checking quality for stacking pins and avoid new problems on the library.
Thanks!

poeschlr:
I would assume the error extra stuff is what would normally be printed only with the -v options. Is this correct?
(meaning the stuff that is reported below a more generic error.)

So might it be better to have one common error that is reported if any of these is violated and keep the current more detailed report as error extra?

diegoherranz:
`errorExtra()` is only shown if using "double verbose" `-vv`.

I guess I could add some `error()` messages and leave the `errorExtra()` ones as they were.
However, the way the code is, I would need to add an `error()` for each `errorExtra()` I have modified since I can't add a generic `error()` above which is common for them all.
Also it seems that we report quite a few `error()` with similar level of detail already:
Examples: https://github.com/KiCad/kicad-library-utils/blob/master/schlib/rules/S4_3.py#L100 or https://github.com/KiCad/kicad-library-utils/blob/master/schlib/rules/S4_3.py#L109

Either way let me know and I'll change it or just fix the merge conflict.

Thanks!

P.S.: Just for reference, when Travis checks the symbol libraries, it runs `comparelibs.py` which runs `checklib.py` using `-vv`. (https://github.com/KiCad/kicad-library-utils/blob/master/schlib/comparelibs.py#L56)

poeschlr:
I am ok with it as long as there is still reasonable output even without the -v options. So please rebase.

diegoherranz:
Rebased to master. Thanks!

poeschlr:
Thanks

PR #269, opened 2018-11-18
schlib/rules: style fixes

diegoherranz:
Most of them as reported by flake8.
Still a few things to be fixed, but this was the most obvious/trivial stuff.

Let me know if you disagree with any change and I can fix or remove it.

Thanks!


poeschlr:
There is a lot of noise in the source code changes as you have one commit for fixing programming style stuff and for adding your changes.

Could you detail what this really changed and where i should look for in the files?

diegoherranz:
It is only coding style. I did split this from https://github.com/KiCad/kicad-library-utils/pull/268 on purpose to make the review easier. And in case we only wanted the actual fixes and not the coding style ones.

poeschlr:
I am not so sure about your changes for the "not in" operators. I feel `x not in y` is easier to read than `not x in y` (The not in operator is a special operator defined to be the exact inverse of in: https://stackoverflow.com/a/54437309)

Even flake suggest it that way: https://lintlyci.github.io/Flake8Rules/rules/E713.html

diegoherranz:
That's what I think I did, turning a few `not x in y` into `x not in y`.
Example: https://github.com/KiCad/kicad-library-utils/pull/269/files#diff-18755e901d1e91a969f0d39e9bc19e27L66

poeschlr:
I am an idiot. I read the diff wrong. (don't ask why, i don't know how i managed that.)
Ignore my last response. All is well. Thanks

PR #271, opened 2018-12-10
Minor improvements to sch + schlib

madworx:
sch + schlib: Add __init__py. files to enable kicad-library-utils to be included as a Python module.
schlib: Enable finding modules by "alias" in addition to name.


poeschlr:
are the init files empty on purpose?

poeschlr:
Empty init files are not needed it seems for python 3: https://stackoverflow.com/a/48804718 So i would suggest to fill out the init files with importing sub files as required. That allows you to import the library without needing to know how the files are named inside it.

Python 2 is dead so we really do not want to sink any resources in supporting it ;) https://pythonclock.org/

PR #272, opened 2018-12-11
extends the range of Conn_2Rows to 75 pins 

herostrat:
Reason is preparation of https://github.com/KiCad/kicad-footprints/pull/879

It adds 3 new connectors.
The previous limit was afaik arbitrary.


herostrat:
I now have, sorry for the huge delay.
https://github.com/KiCad/kicad-symbols/pull/1825

myfreescalewebpage:
Symbols have been merged.

PR #273, opened 2018-12-13
G1.7 should produce an error, not a warning

fauxpark:
The [wording of this rule](http://kicad-pcb.org/libraries/klc/G1.7/) "must always" suggests that this check should error out if it encounters a file with CRLF line endings. The .gitattributes file should catch most of these by automatically converting to LF during the commit, but apparently it does not apply when committing through the browser.

The unfortunate part is that when working on Windows, running the check before committing will show this error, but provided there are no other issues flagged you can just commit anyway and reset your working tree.


poeschlr:
Thanks

Issue #275, opened 2018-12-22
Allow step files in footprints


drid:
According to https://github.com/KiCad/kicad-packages3D both WRL and STEP can be used.

KLC F9.3 should be updated accordingly

poeschlr:
Sorry no we do not allow footprints to point to step files. If kicad needs step it does exchange the file format automatically. We are aware that a better solution would have been to have no file ending at all but we did not have the luxury to start with a clean slate.

PR #276, opened 2018-12-22
#275 Allow step files in F9.3

drid:
According to https://github.com/KiCad/kicad-packages3D both WRL and STEP can be used.

KLC F9.3 should be updated accordingly


herostrat:
Just for better overview
This is the Issue drid opened for this PR:
https://github.com/KiCad/kicad-library-utils/issues/275

fauxpark:
[F9.3.8 states](http://kicad-pcb.org/libraries/klc/F9.3/):

> Filetype must be .wrl (KiCad replaces it with .step automatically for the step export.)

As far as I can tell it's not necessary for the check scripts to care about STEP files since they are apparently only used for MCAD integration. Thus it was actually removed in https://github.com/KiCad/kicad-library-utils/commit/15bec825d90951bba156bd8a24276b88d6ca58d4.

poeschlr:
@fauxpark is correct. i detailed it in the original issue. Closed as invalid.

PR #280, opened 2019-01-12
Clean up DIP switch generator script

fauxpark:
As with #256, this script now generates DIP switch symbols with KiCadSymbolGenerator.

The only differences between the generated symbols and the current library symbols are the out-of-order drawing lines (which KiCad will fix on save) and the deMorgan parameter (the generator defaults to 1).


poeschlr:
Thanks

Issue #278, opened 2019-01-01
Have Travis check for duplicate lines (maybe duplicate pads) 


Shackmeister:
I have seen a few footprints which has multiple lines in the same spots, probably because the were copied and paste and forgot about

cp-aquila:
Can you give an example of a footprint where that applies?

Shackmeister:
For instance this one https://github.com/KiCad/kicad-footprints/pull/1118

Not saying it's a huge problem, but it could be nice to catch them when they come

myfreescalewebpage:
Already seen too and not so easy to catch when making a review!

cpresser:
New patch, that checks different layers.

PR #283, opened 2019-01-18
Check courtyard layers for duplicate lines

cpresser:
Related: https://github.com/KiCad/kicad-library-utils/issues/278

This code is rather ugly; the nested loops to check overlap could be implemented that no combination is checked twice.
But well, it works for me


cpresser:
Closing this PR.
The original issue did not ask for courtyard, but for lines on all layers.
I will redo this patch so it works on different layers.

Besides, #284 also kind of solves the issue of duplicate lines for the courtyard. If there is a duplicate line on the courtyard layer, the patch from #284 also displays that as an error.

Issue #277, opened 2019-01-01
Have Travis check courtyard continuity 


Shackmeister:
It would be a great addition if we could have the script check if the courtyard is closed

cpresser:
This should be the patch you are looking for :)

poeschlr:
I would assume this is done. Reopen if i am wrong.

PR #284, opened 2019-01-25
Check if courtyard is closed

cpresser:
Related #277

### DIFF
--- a/pcb/rules/F5_3.py
+++ b/pcb/rules/F5_3.py
@@ -99,7 +99,7 @@ class Rule(KLCRule):
 
     def isClosed(self, layer):
       def isSame(p1, p2):
-        s = p1['x'] == p2['x'] and p1['y'] == p2['y']
+        s = abs(p1['x'] - p2['x']) < 0.00001 and abs(p1['y'] - p2['y']) < 0.00001
         return s
 
       # no line is considered as closed
--- a/pcb/rules/F5_3.py
+++ b/pcb/rules/F5_3.py
@@ -226,8 +226,8 @@ class Rule(KLCRule):
 
         # Check that courtyard is closed
         if len(self.unconnected) > 0:
-            self.error("Courtyard should be closed.")
-            self.errorExtra("The following lines have unconnected enpoints")
+            self.error("Courtyard must be closed.")
+            self.errorExtra("The following lines have unconnected endpoints")
             for bad in self.unconnected:
                 self.errorExtra(graphItemString(bad, layer=True, width=False))
 
--- a/pcb/rules/F5_3.py
+++ b/pcb/rules/F5_3.py
@@ -96,6 +96,47 @@ class Rule(KLCRule):
         else:
             return None
 
+
+    def isClosed(self, layer):
+      def isSame(p1, p2):
+        s = p1['x'] == p2['x'] and p1['y'] == p2['y']
+        return s
+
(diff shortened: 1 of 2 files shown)


JacobEFO:
Hi @cpresser 

I for the fun of it fiddled around with your PR here. Your solution seems to be working well with square courtyards:
Take for instance: DFN-10-1EP_2x3mm_P0.5mm_EP0.64x2.4mm
image
With a corresponding check_kicad_mod.py output: 
image
When we open the courtyard:
image

But, if we give it a round courtyard:
image
The check_kicad_mod output becomes:
image

If arcs with a 360 degree turn-around is used, the corresponding output is still "Courtyard should be closed".

An additional feature could maybe be, if it returns the coordinates at/around which the error is generated?

cpresser:
Hi @JacobEFO 
> If arcs with a 360 degree turn-around is used, the corresponding output is still "Courtyard should be closed".

The lastest commit should fix that. For arc, the start-point was not extracted correctly.
Now it does calculate the actual start point.
I have tested it with:
* two arcs forming a circle
* two arcs and two lines
* lots of lines and one arc
* a single circke

> An additional feature could maybe be, if it returns the coordinates at/around which the error is generated?

Actually, thats already there. use "-vv" to show the extraErrors. They contain the line/arc that failed the check

JacobEFO:
Hi @cpresser

Btw I am not a librarian, I just randomly checked out your implementation and found the error.

Seems like you did test it thoroughly though.

Ah, apparently -v does not catch all the errors. Lol. Good call.

poeschlr:
For testcases regarding courtyards with arc sections look in the battary holder library.

poeschlr:
Something to keep in mind for arcs: The kicad file format is a bit limited. It is nearly impossible to create precisely closed outlines when arcs get involved (unless you have them only with n*90 degrees)
This is the reason why kicad allows for some tolerance to count an outline as closed (I think the tolerance is 0.01mm.)
Meaning we might want to have this tolerance allowance also for this check. (But i would only allow it for arcs.)

cpresser:
Force push, I did test with 0.1mm and left that value in. Fixed with the force push.

poeschlr:
0.1mm is a bit large if kicad uses 0.01 ;)

cpresser:
it is set to 0.01mm with the last commit. See line 121 in the changed file. 
The comment was not that clear, sorry.

Shackmeister:
@cpresser @poeschlr 
Any remarks on this?
from my testing it seems to work fine

cpresser:
right now I consider this done. I am not aware of any issues.

pcb/rules/F5_3.py
```diff
@@ -174,9 +224,17 @@ def check(self):
             for bad in self.bad_grid:
                 self.errorExtra(graphItemString(bad, layer=True, width=False))
 
+        # Check that courtyard is closed
+        if len(self.unconnected) > 0:
+            self.error("Courtyard should be closed.")
```
poeschlr: Wrong wording. Use: Courtyard **must** be closed. (emphasis to make it clear what word to change.)

cpresser:
Fixed the wording and also found another typo.

poeschlr:
Thanks. Lets see how this goes down if we let it loose to the wild. (I suspect we will find issues as this is a quite complex addition but i assume it to work for the most part.)

PR #286, opened 2019-01-25
Check for duplicate lines

cpresser:
Related #278


cpresser:
This checks for duplicate/overlapping lines in silkscreen and fab layers.
Courtyard layers are checked by #284 and not included in this patch.

poeschlr:
Merging this would require documenting this also on the website. Would you be up to extending rule 5.1 with this? We can not include any check against undocumented rules ;)

cpresser:
Imho it would require changes in 5.1 and 5.2
And yes, I can provide a patch for KLC.
But I can't seem to find the respository that has the KLC souce.

evanshultz:
https://github.com/KiCad/kicad-website/tree/master/content/klc

cpresser:
I just figured that this patch could be considered as incomplete.
It only works for lines, but not arcs and circles.
Circles are simple; arcs could be done, there is similar logic in #284 that I could reuse.

What do you think?

poeschlr:
If we extend this to courtyard (and other layers) then we might want to write a new separate rule that clarifies this instead of adding this to every rule.

This would then mean that this check should also be in a separate file.

cpresser:
I made another patch to check the courtyard (#284). That's why I did not include the courtyard check in this patch.
I was about to postulate that this is already enough. But there are cases where this is not true. A rectangular courtyard which is fully duplicated (8 lines total) will be considered as closed. Right now I am lacking ideas on how to change #284 to detect that as well.

That means this patch should also check the courtyard layers. Adding a check for circles is trivial; a update for this PR will be ready soon.

Making a F5.4 rule that checks all layers seems like a good idea.
I started a pad to coordinate the text: https://pads.ccc.ac/p/KLC_F5.4

cpresser:
A changed patch has been (force) pushed. 
It now checks for duplicate/overlapping lines and for duplicate circles.
I have some Ideas for Arcs as well, but would like to do that in a separate PR.

KLC (website) has a PR as well: https://github.com/KiCad/kicad-website/pull/384

poeschlr:
I think this is ok now. lets test it on the real world.

poeschlr:
@cpresser it seems your script has false positives if a two lines meet at a point and are in the same angle (so if a long line is split into two smaller lines.)

cpresser:
@poeschlr Good catch.
This is a case were the check failed.
A patch to fix this is pretty simple. A PR is here: https://github.com/KiCad/kicad-library-utils/pull/301

PR #287, opened 2019-02-06
Better wording for missing 3d model settings

poeschlr:
I noticed many users think they must provide a 3d model. The sparse wording of the travis error message might be partly to blame for that.

I think this change will make it more clear to users what we really want from a footprint.

### DIFF
--- a/pcb/rules/F9_3.py
+++ b/pcb/rules/F9_3.py
@@ -172,18 +172,18 @@ class Rule(KLCRule):
             # Warning msg
             if module.attribute=='virtual':
                 # virtual components don't need a 3D model
-                self.warning("No 3D model settings provided for virtual component")
+                self.warning("Optional 3D model path missing from the 3D model settings of the virtual footprint")
                 self.no3DModel = True
                 return False
             else:
-                self.error("Footprint 3D model settings are not allowed to be empty")
+                self.error("3D model path missing from the 3D model settings of the footprint")
                 self.no3DModel = True
                 return True
 
         self.tooMany3DModel = False
         if len(models) > 1:
             self.tooMany3DModel = True
-            self.warning("More than one 3D model setting provided")
+            self.warning("More than one 3D model path provided within the 3D model settings of the footprint")
 
         model_error = False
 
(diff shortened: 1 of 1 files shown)


evanshultz:
Makes sense. Rather than the word "settings" what about "reference"?

poeschlr:
I am not sure if reference would be the right word. https://en.oxforddictionaries.com/definition/reference

Maybe "link"? Or "path"?

diegoherranz:
+1 for "path"
I don't find "settings" very clear.

poeschlr:
Updated wording

evanshultz:
I think the first description in the link you included above is perfect. This section of the footprint is mentioning (referring to, hence "reference") the 3D model.

Path, in my mind, is the journey but not the destination. In this context, I see "path" as the series of folders but not the file (the 3D model here). So "path" doesn't sound right to me either.

No other suitable words are coming to me. Any other suggestions? Or maybe my suggestion was bad?

poeschlr:
On file systems the path is normally the full thing including the file name. Example the python path module provides many functions that indicate that the path includes the filename: http://www.tutorialspoint.com/python/os_path_methods.htm

poeschlr:
I changed it to use "file path" now which should make it really clear what is meant.

evanshultz:
I see now why you may have went with "settings" at first: the UI uses that term. It doesn't sound too awkward in Pcbnew, though there's probably a better word. Just a note to let you know that I now better understand where you were coming from.

The wording is cumbersome to me. Lots of words but still not super clear. I'll hold off a bit from making any suggestions, unless asked, to see if there is feedback from anybody else.

Also note that 5.1 is intended to have multiple models for one footprint, so perhaps we don't need that case?

poeschlr:
I know the wording is not perfect, but it is better as it was in the past. Unless someone can come up with a better wording then this might be best merged.

---

Regarding multiple 3d models: I would not really use multiple models at this point in time for the official lib. The vast majority (if not all) footprints are best served by a single model. So the check is still a valid sanity check. (Multiple models where always supported. The new thing added is that one can control which one is visible.)
It is a warning anyways so i think it really should stay in.

Shackmeister:
The multiple model is a pretty big misunderstanding. It's only the preview you can disable. It will still be visible in the board. It's only meant to be able to align multiple 3d models on the same

evanshultz:
OK. Will merge.

I was going off of Nick's comments in a Launchpad bug about multiple models and hadn't tried it myself. Thanks for clearing things up, Frank.

PR #295, opened 2019-05-13
Use option flag to decide if symbol is a power symbol

seppestas:
This fixes the PWR_FLAG violating KLC rule S6.2 "Ref(erence) field must be VISIBLE" because it does not use the #PWR reference.


poeschlr:
Thanks

PR #296, opened 2019-05-13
Add exclude option to checklib to exclude particular rules

seppestas:
This allows to exclude rules checklib.py should not check against. In a lot of cases this makes more sense than just enabling a subset of rules, since some rules might not fit certain libraries.

E.g a lot of symbols in the mainline MCU_Microchip_ATmega.lib library violate S4.4 and almost all symbols in MCU_ST_STM32F4.lib violate S3.1.

This could be easily ignored by running

```
checklib.py MCU_Microchip_ATmega.lib -e S4.4 -vv
checklib.py MCU_ST_STM32F4.lib -e S3.1 -vv
```

Allowing to ignore the "noise", making it easier to spot new mistakes that might actually matter.


seppestas:
I think it would be even more useful to exclude certain subsets of rules, e.g just ignoring S4.4.1 would only ignore "Power and ground pins should be set to either Power Input or Power Output" and still check the other "subrules". Sadly I did not find a way to do this without greatly overhauling checklib.py.

poeschlr:
good idea for local runs, will not fix travis complaining as there is no way to tell travis which rules to ignore that way.

one option would be to have a library dependent travis setup either in a central config file or as an exclude file added to the library (the later only works in library file formats where the library is a directory)

PR #297, opened 2019-05-14
Allow checking multiple specific components using CSV

seppestas:


poeschlr:
I think argparser has an option to do this aromatically. check the nargs argument.

PR #298, opened 2019-05-14
Seperate LibTable class from check_lib_table functionality

seppestas:
This makes re-using the LibTable class a bit easier and logical. It also remove schematic lib specific code from LibTable making it re-usable for footprint libraries.

Could be used to e.g check custom options in LibTables. I plan to use it together with the functionality of #297 to only check "approved" symbols in a symbol library by having a CSV "SYMBOLS" field in the library options. This makes something like this possible:

```python
# Check "enabled" symbols of all symbol libraries in a symbol library table for KLC compliance
import subprocess
import re
from importlib import import_module
LibTable = import_module("kicad-library-utils.lib_table").LibTable

lib_table = LibTable('sym_lib_table')

for library in lib_table.entries:
	re_symbols= re.search(r'SYMBOLS=([^ ]+)', library['opt'])
	checklib_args = []
	if re_symbols:
		symbols = re_symbols.groups()[0]
		checklib_args.extend(['-c', symbols])

	checklib_args.append(library['uri'].replace('${KICAD_SYMBOL_DIR}', '../symbols'))
	# TODO: make checklib work as python module
	subprocess.call(['./kicad-library-utils/schlib/checklib.py'] + checklib_args)
```

This could also be used to add an option to ignore certain rules for a symbol library, see #296.

### DIFF
--- a/check_lib_table.py
+++ b/check_lib_table.py
@@ -14,100 +14,62 @@ It is important that the official libraries match the entries in these tables.
 from __future__ import print_function
 
 import argparse
-import re
 import os
 import sys
 
-parser = argparse.ArgumentParser(description='Compare a sym_lib_table file against a list of .lib library files')
+from lib_table import LibTable
+
+parser = argparse.ArgumentParser(description='Compare a sym-lib-table file against a list of .lib library files')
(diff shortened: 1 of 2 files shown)


poeschlr:
thanks

PR #301, opened 2019-05-31
Fix F5_3 false positives for colinear lines

cpresser:
Two lines that are colinear and have exactly one identical point are not overlapping.


poeschlr:
Seems to work. Thanks

PR #304, opened 2019-06-03
Add support to compare 2 folders of .lib files

lukh:
--new and --old args can be either .lib files or directories.
In the case they are directories, each .lib file of these dirs are added to the new_libs and old_libs dict.


poeschlr:
thanks

PR #311, opened 2019-08-18
SymbolGenerator: fix mirror of arc objects

lorem-ipsum:
only the arcs direction was mirrored, but not the location of the arc


lorem-ipsum:
I've added another trivial bug fix occurring in the error handling when an alias is already existing

poeschlr:
Thanks

PR #312, opened 2019-10-06
Update metadatacheck to fit v5.1

poeschlr:


schlib/rules/S6_2.py
```diff
@@ -93,12 +93,69 @@ def checkDatasheet(self):
...
 
-    def check(self):
+    def checkDocumentation(self, name, documentation, alias=False, isGraphicOrPowerSymbol=False):
+
+        errors = []
+        warnings = []
+
+        if not documentation:
+            errors.append("Missing all metadata information in the .dcm file (datasheet, keyword and description)")
+        elif (not documentation['description'] or
```
evanshultz: Can lines 108 to 115 be combined and just check for the description or the keywords with another clause at the end for datasheet and power symbols?

schlib/rules/S6_2.py
```diff
@@ -93,12 +93,69 @@ def checkDatasheet(self):
...
+        if desc and name.lower() in desc.lower():
+            warnings.append("Symbol name should not be included in description")
+
+        # Datasheet field should look like a a datasheet
+        ds = documentation.get('datasheet', '')
+
+        if ds and len(ds) > 2:
+            link = False
+            links = ['http', 'www', 'ftp']
+            if any([ds.startswith(i) for i in links]):
```
evanshultz: Perhaps this could be more robust. I'm sure there are good regex for a URL online. Some things I can think of now:
1. `www2` instead of just `www` or `https` instead of just `http`. This will currently work because the starting characters are the same, but it's so rudimentary.
1. Check for `://` or other URL formatting.
1. `html` instead of just `htm`.

And there could be a warning if the link is to a webpage instead of a PDF file. Sometimes that's the best way but the vast majority of our links are PDFs.

cpresser: How about actually fetching the URL?
* Throw an error if the final http status code is not `200 OK`
* Check filetype (perhaps with `python-magic`)

chschlue: Not sure if having the script fetch and process whatever URL random people on the internet throw at it is a good idea.

poeschlr: Can be done in a later pull request. This request will only combine the checks of two rules and align the error messages with the current kicad gui

schlib/rules/S6_2.py
poeschlr: The intention is to combine the two rules with as little changes as possible in this pull request. Any further changes can be made in a separate pull request later.

evanshultz: It looks like all this code is new so I'm not quite following you, but OK. Seems logically correct even if it's verbose and/or hard to follow.

schlib/rules/S6_2.py
```diff
@@ -134,6 +191,35 @@ def check(self):
...
+            invalid_documentation += 1
+
+        # check all its aliases too
+        if self.component.aliases:
+            invalid = []
+            for alias in self.component.aliases.keys():
+                if self.checkDocumentation(alias, self.component.aliases[alias], True, self.component.isGraphicSymbol() or self.component.isPowerSymbol()):
+                    invalid_documentation += 1
+
+        return invalid_documentation > 0
```
evanshultz: No real reason to keep a counter since the result ultimately goes to a Boolean. But I guess in the future it could be handy and it works now too.

evanshultz:
I assume at https://github.com/KiCad/kicad-website/pull/446#issuecomment-575893024 you are referring to this script update?

`check_lib_file` requires at least four pieces of documentation in the lib file. Isn't that wrong? Don't we want F4 to always be blank, and we also accept F3 to be blank for generic symbols that only have a bunch of FP filters? Maybe I didn't quite understand this.

schlib/rules/S6_2.py
poeschlr: All other scripts use counters

schlib/rules/S6_2.py
poeschlr: It is copied from another rule (rule 6.3 no longer exists so all the checks from that file were move in here as these two rules are combined now in the KLC. This is the downside of naming the rule check files after the rule number instead of giving them a representative name of what they actually chekc)

poeschlr:
I now ran the new script and old script on the current master branch of the symbol lib. There is not really a difference in the reports other than what is expected from combining the rules (S6.3 changed to S6.2, detailed messages changed and if both s6.3 and S6.2 is violated then they are of course under the same header which makes diff think something was added in the original)
The logs i created in this process:
[before_change_1.txt](https://github.com/KiCad/kicad-library-utils/files/4087866/before_change_1.txt)
[after_change_1.txt](https://github.com/KiCad/kicad-library-utils/files/4087868/after_change_1.txt)

PR #313, opened 2019-10-25
Skip non-footprint files in check_all script for travis ci

ottojo:
This fixes KiCad/kicad-footprints#1545


poeschlr:
Might it be better to explicitly check for the fp-lib-table such that we still get a warning if there is a file added that should not be there?

ottojo:
That sounds useful indeed, i changed the script accordingly

poeschlr:
thanks

PR #319, opened 2019-11-27
Skip THT checks for SMD components

drid:
SMD components can have pads for thermal vias that should not conform to through hole pad requirements.

Rule F7.4 should be skipped for such components

### DIFF
--- a/pcb/rules/F7_4.py
+++ b/pcb/rules/F7_4.py
@@ -64,7 +64,8 @@ class Rule(KLCRule):
             * pin1_count
         """
         module = self.module
-
+        if module.attribute != 'pth':
+            return
         return any([
             self.checkPads(module.filterPads('thru_hole'))
             ])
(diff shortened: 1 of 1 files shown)


poeschlr:
I would say https://github.com/KiCad/kicad-library-utils/pull/288 is the better approach

drid:
> I would say #288 is the better approach

I agree, initially i started looking for a way to detect a th pad within an smd pad.

My logic behind this PR was that TH pads in an SMD component are always thermal via otherwise it would just be a TH component.

drid:
On second thought this does not cover a TH component with TH pads and smd pads with thermal vias

poeschlr:
The more powerful implementation has been merged. It is still limited but an improvement over what we have now. I documented the limitations and a proposed improved implementation in #322

Feel free to comment on that issue with your own ideas.

PR #320, opened 2019-11-30
Add a generator for capacitor networks.

Doeme:
This is basically just a copy of the R_Networks code, with some minor (i.e. drawing capacitors instead of resistors) and sedding s/resistor/capacitor/g

It is still lacking the proper footprints though.

### DIFF
--- a/schlib/autogen/C_Network/C_Network.py
+++ b/schlib/autogen/C_Network/C_Network.py
@@ -1,0 +1,274 @@
+#!/usr/bin/env python3
+
+import math
+import os
+import sys
+
+sys.path.append(os.path.join(sys.path[0], '..'))
+
+from KiCadSymbolGenerator import *
+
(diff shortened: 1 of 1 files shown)


poeschlr:
Is there a symbol pull request?

Doeme:
Alas, there is not, since I have not figured out how to auto-generate the footprints yet (and for the project that spawned this pull request I hand-crafted the one needed).

I've found some kicad-integrated generators (reached by the footprint-editor), should these be used?

poeschlr:
https://github.com/pointhi/kicad-footprint-generator

Doeme:
Oh, ok, cool. I can at some point include the C_ARRAY footprints there, but at the moment I'm rather preoccupied. It'll probably be 2 months until I work on it again.
If somebody wants to step in: feel free! ;)

Issue #321, opened 2019-12-29
S6.3.2 check for space-separated keywords


aewallin:
checking that symbol keywords are separated by spaces should be quite straightforward?

an automated check would reduce manual work for reviewers (if e.g. comma-separated keywords or something else are caught automatically by travis)

JacobEFO:
I'd love to make a little something for this issue. If you're open for it.

PR #323, opened 2020-03-07
Solves Issue #321: Check for space separated keywords

JacobEFO:
A naive approach to handling space separated keywords fixing #321 

The thought was, that it would be more interesting to find keywords separated with a list of illegal symbols, such as: `,.:;?!<>`

Currently allowing especially hyphen (-) to get through, because it appears to be used already within keywords. A primary issue atm is, that some symbols such as within CPLD_Xilinx.lib contain no "K" field in its .dcm file. This triggered an error, that I handle with a try/except.

if the error is triggered it appends to the warnings, that symbols appear to not have any keywords. If illegal characters are found within the keywords, an error is appended saying it found the following illegal characters and prints them.

I would like to hear, what you think of the solution, and what specific sort of characters should be searched for within the keywords.

### DIFF
--- a/schlib/rules/S6_3.py
+++ b/schlib/rules/S6_3.py
@@ -76,10 +76,13 @@ class Rule(KLCRule):
 
         # Handle keyword seperator. Right now it filters based on a regular
         # expression based on certain symbols. Right now hyphen (-) is still allowed.
-        keywords = documentation.get('keywords', '')
-        forbidden_matches = re.findall('[,.:;?!<>]', keywords)
-        if forbidden_matches:
-            errors.append("Symbol keywords contain forbidden symbols: {}".format(forbidden_matches))
+        try:
+            keywords = documentation.get('keywords', '')
+            forbidden_matches = re.findall('[,.:;?!<>]', keywords)
+            if forbidden_matches:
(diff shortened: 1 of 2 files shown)


cpresser:
I think we need to discuss the list of illegal symbols.
I think the dot `.` should be allowed, for example in `3.3V LDO` or `66.6MHz clock`

JacobEFO:
Hey @cpresser 

The list of symbols was merely a suggestion/prototype. The list is not exhaustive. There may be other more interesting symbols to filter from or some to exclude.
It is merely just a suggestion as to symbols people may end up using instead of a space for seperation of keywords. I don't necessarily expect anyone to use "!?<>" but was just used for testing.

Realistic symbols may be more: ",.:;|" or similarly. Using this list of illegal symbols is just a simple yet efficient way to handle keyword seperation. 
It would be hard to detect someone not using any sort of separator, say: "lowvoltageopamp" would be cumbersome to check for, whereas assuming someone may end up writing "lowvoltage,opamp" is reasonable.

cpresser:
I did a little testing and think that `.` should not be in that list.
There is the use-case of things like `3.3V` Perhaps you can write a re that allows the dot character only if it is followed by a word character.

Missing keywords generating a warning is good. I like that.

PR #324, opened 2020-04-16
schlib improvement to handle double-quote text in symbol user fields

Cisco25:
While working on my little tool to parse [symbol library files](https://github.com/Cisco25/kicad-database-utils), I've noticed that there was a little bug to handle user fields (F3 and up) when a double-quote was inserted in the text field, the library parser was handling it as an escape quote. It appears to come from the shlex command parser which has no insight on how to treat this character.

I've tried to solve it using shlex but eventually gave up, I even posted on StackOverflow:
https://stackoverflow.com/questions/60877782/python-shlex-posix-usage-dilemma

Instead, I've found regex pattern matching for user fields containing double-quotes works well, based on this answer:
https://stackoverflow.com/a/16710842/12794913

I've tested this change quite extensively as I've been doing a lot of library clean-up for my company.
I'm open to other implementation, just wanted to give a heads-up on this little annoying bug.

Here is a little code snippet if you want to understand the issue I'm referring to:
```
import shlex

string = '\"This is a \\"difficult\\" problem\" please help'
print(f'string = {string}')

s1 = shlex.shlex(string, posix=False)
slist1 = list(s1)
# Not the expected result
print(f'slist1 = {slist1}')

s2 = shlex.shlex(string, posix=True)
slist2 = list(s2)
# Not the expected result
print(f'slist2 = {slist2}')

import re

slist3 = re.findall(r'(?:[^\s,"]|"(?:\\.|[^"])*")+', string)
# It works, yay!
print(f'slist3 = {slist3}')
```


schlib/schlib.py
```diff
@@ -153,11 +153,15 @@ def __init__(self, data, comments, filename, documentation):
...
             checksum_data += line.strip()
             line = line.replace('\n', '')
-            s = shlex.shlex(line)
-            s.whitespace_split = True
-            s.commenters = ''
-            s.quotes = '"'
-            line = list(s)
+            if '\\"' in line:
+                import re
+                line = re.findall(r'(?:[^\s,"]|"(?:\\.|[^"])*")+', line)
```
Cisco25, line 158: If double-quote detected in line, split using regex instead of posix

PR #325, opened 2020-04-23
R network split

evanshultz:
Expand the script to make split (each resistor in its own unit) versions of all relevant resistor generators currently in the script.

Symbol PR:


evanshultz:
Symbols have been merged.

vsergeev:
:+1: Found this after looking for `R_Pack04_US`. I noticed that the US style is currently only generated for the `R_Network*` variety. It would be awesome to pick up the US styles for the `R_Pack*` symbols too, if possible. I would contribute a patch here, but after taking a closer look it looks like the body generation would have to be factored out and probably affect all of the generate functions, which is a bit beyond a simple patch.

PR #326, opened 2020-04-24
Update STM32 datasheets

evanshultz:
Update the path of ST datasheets and add what MPNs they apply to, helping out to identify if symbols are scripted such as when encountering https://github.com/KiCad/kicad-symbols/pull/2463.


rleh:
The list of datasheets is unfortunately somewhat old and therefore incomplete. Complete series like STM32G4 are missing.
It is possible to extract a list of all STM32 datasheets from STs website with only a few lines of python code. PR coming soon...

PR #327, opened 2020-05-20
STM32 symbol generator

rleh:
* Get list of datasheets from STs website (and hence remove the outdated `datasheets.txt` file)
* Use multiprocessing to speed up data mining from PDF files
* Filter `*.xml` files: Only take `STM*.xml` files (`compatibility.xml`, etc. cause errors)


rleh:
This somehow supersedes #326
cc @evanshultz

myfreescalewebpage:
@evanshultz do somebody check for this ? Required for https://github.com/KiCad/kicad-symbols/pull/2463

PR #331, opened 2020-08-24
Fix call to checklib.py from comparelibs.py (--nocolor arg and python3)

justyn:
Fixing 2 bugs in comparelibs.py when it called checklib.py:

* Using --nocolor along with extra arguments (eg --footprints) would lead to concatenation, and would therefore fail with eg:
`checklib.py: error: unrecognized arguments: --nocolor--footprints`

* checklib.py was still called with `python` by comparelibs.py instead of `python3`, leading to `python: not found` if `python` is not available on the system, as is the case for Ubuntu 20.04. Availability of python3 is already enforced in both files by their shebangs.


justyn:
> The first part is pretty clear, but how does the shebang enforce python3 availability if the script is invoked as an argument to the interpreter instead of directly?

You're right, sloppy language on my part - I really meant that python3 was already _assumed_ due to the shebang.

I assume from your comment you're trying to prevent breakage for people using python 2?

justyn:
I can update to use `sys.executable` in place of python, that should be the most robust I think?

chschlue:
Py2 of course deserves to rest in peace, but then, why break it for people stuck with it for some reason?
Your proposal looks good to me.

justyn:
Sure, I've pushed the update.

Of course, one day hopefully checklib will be used as a module instead :-)

PR #332, opened 2020-08-26
Din41612 symbols

evanshultz:
This is one of the tasks I wanted to get done for v6 (https://github.com/KiCad/kicad-symbols/issues/2705#issuecomment-629806378), and this part of it turned out to be easy enough that I will try to sneak it in now.

Triple-row symbols will take some framework, err, work, but single- and double-row symbols could be supported with the generous support already available in the script. This supports most of the DIN41612 footprints so I felt taking more time and changing a lot more stuff was a case of 'letting perfection be the enemy of good' if this can get done quickly.

Symbol PR: https://github.com/KiCad/kicad-symbols/pull/2935


evanshultz:
@myfreescalewebpage 
Can you merge this? You merged the symbols that came from this generator so I assume you're OK with this.

myfreescalewebpage:
Sure, sorry I have miss to do it !

PR #334, opened 2020-09-05
Add rouding to mmToMicrons() function

### LINKED ISSUE #329: F5.3 and checking non-line elements
I was wondering why the F5.3 check at KiCad/kicad-footprints#2267 failed. The problem here is the check expects the elements to be line elements only, while the footprint contains arc elements on the CrtYd layer. Which is against the KLC, I guess ("The [...] courtyard is defined as the [...] **rectangular area** [...]. It is allowed to create a contoured courtyard area using a **polygon** instead of a simple rectangle.")? On the other hand the check does not complain about having non-line elements on the CrtYd in the first place.

So what's your opinion about this? Possible solutions could be:
- ignoring non-line elements in the code and/or
- adding a check to enforce line elements only

cpresser:
For when using getStartPoint() on arcs combined with the mmToMicrons()
function the accuracy was not good leading to errors of the F5.3 check.
See https://github.com/KiCad/kicad-footprints/pull/2267

With rounding applied to the value before converting it to int that issue
is gone. This change won't break other rules since the function mmToMicrons()
is only used for rule F5.3.

Fixes #329

### DIFF
--- a/pcb/rules/rule.py
+++ b/pcb/rules/rule.py
@@ -23,7 +23,7 @@ def mmToMicrons(mm):
     elif mm > 0:
         mm += 0.0000005
 
-    return int(mm * 1E6)
+    return int(round(mm * 1E6))
 
 def getStartPoint(graph):
     if 'center' in graph:


evanshultz:
OK. Looks simple enough. Thanks!

cpresser:
> is gone. This change won't break other rules since the function mmToMicrons()
> is only used for rule F5.3.
> 
> Fixes #329

Well... turns out my testing was not good enough:
image

Will push a fix for that soon.

PR #336, opened 2020-09-13
Handle module name as string

akosmarton:
If the module name contains only numeric characters, the variable should be converted to string to ensure the proper rules checking.

### DIFF
--- a/pcb/rules/F5_2.py
+++ b/pcb/rules/F5_2.py
@@ -29,7 +29,7 @@ class Rule(KLCRule):
             self.error("Missing 'value' field")
             return True
 
-        if not val['value'] == mod.name:
+        if not str(val['value']) == mod.name:
             errors.append("Value text should match footprint name:")
             errors.append("Value text is '{v}', expected: '{n}'".format(
                 v = val['value'],
--- a/pcb/rules/F9_1.py
+++ b/pcb/rules/F9_1.py
@@ -49,7 +49,7 @@ class Rule(KLCRule):
             self.error("footprint name (in file) was '{0}', but expected (from filename) '{1}'.\n".format(module.name, os.path.splitext(os.path.basename(module.filename))[0]))
             err = True
 
-        if module.value['value'] != module.name:
+        if str(module.value['value']) != module.name:
             self.error("Value label '{lbl}' does not match filename '{fn}'".format(
                 lbl=module.value['value'],
                 fn = module.name))
--- a/pcb/kicad_mod.py
+++ b/pcb/kicad_mod.py
@@ -62,7 +62,7 @@ class KicadMod(object):
         self.sexpr_data = sexpr_data
 
         # module name
-        self.name = self.sexpr_data[1]
+        self.name = str(self.sexpr_data[1])
 
         # module layer
         self.layer = self._getValue('layer', 'pth', 2)


pcb/rules/F5_2.py
```diff
@@ -29,7 +29,7 @@ def checkMissingValue(self):
             self.error("Missing 'value' field")
             return True
 
-        if not val['value'] == mod.name:
+        if not str(val['value']) == mod.name:
             errors.append("Value text should match footprint name:")
             errors.append("Value text is '{v}', expected: '{n}'".format(
                 v = val['value'],
```
cpresser, line 35: also convert to str here.

pcb/rules/F9_1.py
```diff
@@ -49,7 +49,7 @@ def check(self):
             self.error("footprint name (in file) was '{0}', but expected (from filename) '{1}'.\n".format(module.name, os.path.splitext(os.path.basename(module.filename))[0]))
             err = True
 
-        if module.value['value'] != module.name:
+        if str(module.value['value']) != module.name:
             self.error("Value label '{lbl}' does not match filename '{fn}'".format(
                 lbl=module.value['value'],
```
cpresser, line 54: same as above.

cpresser (commented):
This looks simple enough.
But I have no clue why this change is needed. Can you elaborate on that?
Do you have an example footprint where this patch changes the outcome of the check script?

pcb/rules/F9_1.py
akosmarton: For example Molex uses numeric-only part numbers.

image

You can see the error message below if the module name contains only numeric characters:

```sh
./check_kicad_mod.py ~/work/sharkrf-kicad-library/footprints/srf_Molex.pretty/5051100892.kicad_mod
```
```python
  File "./check_kicad_mod.py", line 117, in <module>
    rule.check()
  File "/home/makos/work/kicad-library-utils/pcb/rules/F7_4.py", line 140, in check
    return any([self.checkPads(module.pads)])
  File "/home/makos/work/kicad-library-utils/pcb/rules/F7_4.py", line 46, in checkPads
    exposed_pad_search = re.search('\-\d+[EP]{2}', fpName, re.IGNORECASE)
  File "/usr/lib64/python3.8/re.py", line 201, in search
    return _compile(pattern, flags).search(string)
TypeError: expected string or bytes-like object
```

evanshultz:
That is not a valid footprint name for this library. See our footprint naming scheme at https://kicad-pcb.org/libraries/klc/F3.6/. And these check scripts are for those guidelines/rules. So as @cpresser said, this doesn't hurt anything but it's not needed when building footprints that conform to the rules of this library.

cpresser:
I think this is a use full addition anyway. The scripts should not throw a cryptic error message just because the user put in stupid input.

The next logical step is to create a `F3.py` that actually checks that rule. I just skimmed over the F3 section of KLC, and it seems there is no way a footprint can only contain numbers. So we could check that.
And perhaps throw a warning if its number of characters is <3.
And throw errors if it contains evil code-points (thinking of emojis, german umlauts, ...)

chschlue:
> german umlauts

Those are _still_ considered evil?
SCNR

chschlue:
Bad jokes aside, we might have to (and be able to) relax naming rules. See https://github.com/KiCad/kicad-symbols/pull/2938

cpresser:
Nevertheless, we can have a script output a warning if non latin chars are used.
A good check would verify that any char is present in the kicad font. Basically an explicit whitelist.


**Somewhat off-topic**
Footprint-names are also file-names. Yay!
I have no clue how the `Pi` would do as filename for windows/linux/mac users. With different locales even more complication could arise.

chschlue:
Yes, I'm not opposed to this PR.

OT: True, the π example is a symbol. But all supported platforms can handle multibyte filenames and assuming there have already been users calling their schematics `MyCool80'sHeävÿMetälPCB.sch` or something without issue, I guess KiCad's codebase isn't too brittle.

Even more OT: I wonder how many ERP systems these π parts broke along the supply chain ;)

PR #337, opened 2020-09-18
Fix F5.3 check

cpresser:
This fixes a regression from #334.

---

My testing when making #334 was not good enough. For some cases the weird `+0.00005` math breaks the rounding really bad.
I remove that block completely, the rounding is simpler and better.
At least my current testing shows that it now works again as intended.

@evanshultz can you please take a look. This breaks travis (false positives). So I would like to address that right away.

See https://github.com/KiCad/kicad-footprints/pull/2438#issuecomment-694928293

### DIFF
--- a/pcb/rules/F5_3.py
+++ b/pcb/rules/F5_3.py
@@ -177,12 +177,14 @@ class Rule(KLCRule):
 
         self.courtyard = self.fCourtyard + self.bCourtyard
 
-        GRID = int(KLC_CRTYD_GRID * 1E6)
 
         # Check for intersecting lines
         self.unconnected.extend(self.isClosed(self.fCourtyard))
         self.unconnected.extend(self.isClosed(self.bCourtyard))
 
+
+        # Check for elements that are not on the grid
+        GRID = mmToNanoMeter(KLC_CRTYD_GRID)
         for graph in self.courtyard:
             if graph['width'] != KLC_CRTYD_WIDTH:
                 self.bad_width.append(graph)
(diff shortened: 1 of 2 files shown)


chschlue:
Two questions:
- What is this function supposed to do? If I'm not totally mistaken, a millimeter is 1e3 microns, not 1e6.
- Isn't the int cast redundant? According to https://docs.python.org/3/library/functions.html#round, round() with a single parameter returns an int anyway.

cpresser:
ad1. To check if lines are on a grid, their coordinates are converted to ~~micrometer~~ nanometer. That should return an integer so a modulo division can be used to determine if they are on the grid.
I agree that 1E3 and 1E6 seems wrong. But it is wrong at two places, so the code works :smile: 
`GRID = int(KLC_CRTYD_GRID * 1E6)`
Proposed fix, rename the function to `mmToNanometer()`

ad2. I was not aware, I will change that.

cpresser:
I did a force-push to keep all changes in one commit. This is so small, adding a 2nd commit seemed wrong.

pcb/rules/F5_3.py
```diff
@@ -194,14 +196,16 @@ def check(self):
...
-            y2 = mmToMicrons(end['y'])
-
+            # make a list of all x and y coordinates of this graphical elements
+            x1 = mmToNanoMeter(start['x'])
+            y1 = mmToNanoMeter(start['y'])
+            x2 = mmToNanoMeter(end['x'])
+            y2 = mmToNanoMeter(end['y'])
             check = [x1, y2, x2, y2]
 
+            # use a modulo division do check if those coordinates are on the grid
```
evanshultz, line 206: change `do` to `to`.

pcb/rules/F5_3.py
```diff
@@ -194,14 +196,16 @@ def check(self):
...
+            # if at least one of the coordinates is not on the grid, add this
+            # element to the bad_grid list
             grid_error = False
             for c in check:
                 if not (c % GRID) == 0:
```
chschlue, line 211: Sorry if I'm just not aware of some Python implicit boolean conversion quirks or I'm not getting the "Python way" of doing things, but this looks like an overly complicated way of expressing `if c % GRID:` to me.
Same (to a lesser extent) with the explicit comparisons of len() to zero below.

cpresser: True, that seems weird. But that is not something I want to address as part of this PR. 
Those checks are were not introduced by me and I don't really feel responsible for them. At least they are out of scope for this PR.
The main goal here is to fix my fuckup of #334 which is breaking travis-ci in production.

I plan to set up some kind of CI for the check-scripts after we migrated this repo to gitlab. After that I am open to cleanup and refactoring.

evanshultz:
I'm fine to do it. Anybody object to merging now as is?

pcb/rules/F5_3.py
chschlue: No problem. I'm just trying to wrap my head around all that Python madness (`not` instead of `!`, `<space>` instead of `<tab>` and whatnot ;) .)

